PR #4, opened 2021-05-07
Add airdrop accounts command

dimiandre:
Procedure

- Export snapshot with `junod export-airdrop-snapshot [airdrop-to-denom] [input-genesis-file] [input-exchange-addresses] [output-snapshot-json] --juno-supply=[juno-genesis-supply]`
- Init genesis with `junod init <moniker> --chain-id juno-1` 
- Import airdrop accounts with `add-airdrop-accounts [airdrop-snapshot-file] [denom]`
- Fix other parameters by hand (genesis time, other accounts, gentxs.. )

### DIFF
--- a/cmd/junod/cmd/airdrop.go
+++ b/cmd/junod/cmd/airdrop.go
@@ -119,7 +119,7 @@ Example:
 				return err
 			}
 
-			// Read Poll file
+			// Read exchanges file
 			exchangeJSON, err := os.Open(exchangeFile)
 			if err != nil {
 				return err
@@ -194,7 +194,7 @@ Example:
 
 				val := validators[delegation.ValidatorAddress.String()]
 
-				// If an account was delegated to an exchange skip
+				// Skip delegations to exchanges
 				if exchangeMap[sdk.AccAddress(delegation.ValidatorAddress.Bytes()).String()] {
 					continue
 				}
@@ -262,10 +262,6 @@ Example:
 
 	return cmd
 }
-
-const (
-	flagGenesisTime     = "genesis-time"
-)
 
 // AddAirdropAccounts Add balances of accounts to genesis, based on cosmos hub snapshot file
 func AddAirdropAccounts()  *cobra.Command {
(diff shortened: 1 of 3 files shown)


faddat:
Hey can we please talk about intent here a little bit?

Let me tell you what I think it looks like and I'd like you to tell me if I get it right. When I read this, I think it looks like an attempt to make it really really easy to export state from Juno so that we can do airdrops, lots of them.

If it's some thing else please let me know.

I think this is really good stuff and I hope start running a validator pretty soon as well. It's really too bad about the whole arm CPU thing but hopefully we can fix that over time.

dimiandre:
> Hey can we please talk about intent here a little bit?
> 
> Let me tell you what I think it looks like and I'd like you to tell me if I get it right. When I read this, I think it looks like an attempt to make it really really easy to export state from Juno so that we can do airdrops, lots of them.
> 
> If it's some thing else please let me know.
> 
> I think this is really good stuff and I hope start running a validator pretty soon as well. It's really too bad about the whole arm CPU thing but hopefully we can fix that over time.

Bascially Juno supply will be based on Cosmos Hub 3 upgrade snapshot, so we need something that can generate the genesis.json based on that. It will be a one time generation, but I think it's good to have in junod binary if we need something similar in the future

blockcreators:
> Thank you and oh boy did I ever misunderstand the general concept here.
> 
> I'm sorry about that.
> 
> I will try and give this a much deeper read...
> 
> TBH, I still have some hesitations about the way that the Juno airdrop is being constructed, though surely after speaking with some people I definitely understood the reasoning.
> 
> I shall restate a little thinking:
> 
> Why not just a one to one airdrop on staked atom as of "x day"?
> 
> .25 to one airdrop on unstaked atom
> 
> Exclude the big evil stealing wallet and any other known thieves

Hey,

So it's 1:1 up to 50k which is the cut off point for whale accounts.
The whale excess goes mostly into a community ran fund, with 10% of the excess going into the dev pool which is held by a group of multisig users that the community voted on. 

All the things were voted on by the community as much as was possible. Exchange validators were excluded.

JakeHartnell (approved):
LGTM. This is much simpler now, and nice work with the code comments.

Note, I've not tested this but will do before we merge Zaki's PR.

faddat (commented):
Once #8 is merged, I'll add this to it locally, and fire it up.  

Thing gives me blurt vibes

:)

JakeHartnell (approved):

dimiandre:
Actually maybe we can keep this in a separate branch, no need to have airdrop commands on main binary since it's just one time generation.

faddat:
@dimiandre that's exactly what we did with Blurt. 

It's in a branch called snapshot in case anyone ever wants to verify. How we started.

Issue #7, opened 2021-05-19
Update Scaffold


faddat:
Starport has been moving fast.  

I'll shortly submit a PR that updates this to the latest scaffold.

blockcreators:
Merged

PR #8, opened 2021-05-19
rebuild with latest starport, move in the Makefile & readme

faddat:


faddat:
Closes #7

.github/workflows/build.yml
```diff
@@ -29,11 +29,11 @@ jobs:
       - name: Compile
         run: |
           go mod download
-          cd cmd/junod
+          cd cmd/Junod
```
jackzampolin, line 32: why `J` instead of `j`

faddat: consistency:

Repo name is:

github.com/CosmosContracts/Juno

faddat: but heh, seems everywhere else, it's lowercase.  

:P

I'll lowercase this one for consistency

:P

faddat:
Here's a proposal for this one, why don't we wait until the final release of Starport 16 and then do this. Let me know what you think about the capitalization because obviously it's quite easy to do either way.

JakeHartnell:
> Here's a proposal for this one, why don't we wait until the final release of Starport 16 and then do this. Let me know what you think about the capitalization because obviously it's quite easy to do either way.

Seems like it's out. : )

Lower case is better please. :pray:

Thanks for this btw! :tada:

faddat:
All set

faddat:
You know last night I probably forgot the address prefix

app/prefix.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	AccountAddressPrefix = "juno"
+	AccountAddressPrefix = "cosmos"
```
JakeHartnell, line 8: Shouldn't this be `juno`?

faddat: Yes

JakeHartnell (approved):
LGTM

PR #13, opened 2021-06-06
Dockerize spike

faddat:
I think this is ready to rock, and if not I'll use CI to debug it, since the build is currently broken anyhow.


the-frey:
@faddat sorry only just saw this comment. Merging now 👍

PR #16, opened 2021-06-14
Fix airdrop

faddat:
just bringing it up to date with main.  

Not sure this is any kind of fix.


faddat:
Hey guys when I wake up I've got to correct "root"--kinda neglected the airdrop module.....

faddat:
Airdrop module is in.  

Goal: run it in CI (it leaves a log that way) and use that as genesis.

the-frey:
I think the path to the binary in the CI might be wrong? That looks like why it's failing the check

faddat:
this is ready to be merged into Zaki's PR.

JakeHartnell (approved):
LGTM

the-frey:
PR is too big to really get my head around. CI builds it though and we need to get it updated, so probably best to merge into airdrop ASAP so we can thrash out any issues that arise

the-frey (approved):

Issue #15, opened 2021-06-10
Generated genesis file doesn't work.


JakeHartnell:
Recent version here: https://github.com/CosmosContracts/testnets/blob/main/juno-testnet-7/genesis.json

When it's run it throws an error:
```
panic: invariant broken: bank: total supply invariant
        sum of accounts coins: 72819737204316ujuno
        supply.Total:          52819737204316ujuno
```

blockcreators:
Denis from discord said:

> When a genesis file is exported the bank module exports supply. In the genesis file above it's an empty array, is there a reason for this? The cosmoshub-4 genesis, for example, has the following value:

```

"supply": [
  {
    "amount": "268335648775167",
    "denom": "uatom"
  }
]
```

faddat:
Good sir block creator, can you please link to the Genesis file that you're referring to using an IPFS CID so that I know exactly what file you're talking about?

For example, are you referring to our current Genesis file or are you referring to the one from the hub?

the-frey:
I assume the one on Github in the testnets repo?

the-frey:
(the one Jake has linked)

faddat:
Yeah, I think I was temporarily blind.  Thanks!

the-frey:
How was this genesis file generated? When was it generated?

the-frey:
The gap between those two numbers is exactly `20000000000000` which makes me think it's not a percentage - so it can't be the e.g. community pool right?

the-frey:
Unless community pool is allocated an _exact number_ in which case I think the fix is likely trivial

the-frey:
Ah no wait, I've got it backwards 'cos accounts coins exceeds supply, my bad.

the-frey:
Okay, maybe try the genesis on this branch? Not sure if `supply` needs to be set as well as the community pool config param, but we can remove if it's not needed https://github.com/CosmosContracts/testnets/tree/community_pool

the-frey:
@JakeHartnell what command were you running when it threw that error? I'll try & replicate and then test this fix locally.

JakeHartnell:
This comment from @dfadeev was helpful: https://github.com/CosmosContracts/Juno/issues/15#issuecomment-860249934

Supply was not being set `"supply": []`

Setting it to:
```
"supply": [
        {
          "denom": "ujuno",
          "amount": "72819737204316"
        }
      ],
 ```
Seems to have done the trick.

[...]

PR #18, opened 2021-06-23
Airdrop command changes

the-frey:
WIP for now, needs more testing

- adds community pool param to `AddAirdropAccounts`
- uses this amount plus a counter of balances to set total supply
- marshals the two above into the output genesis file


the-frey:
Probably https://github.com/CosmosContracts/Juno/pull/16 is worth getting into the `zaki-airdrop` branch before this one is done as well, as there's likely to be some changes needed.

the-frey:
Now #16 is merged into `zaki-airdrop`, I've merged that up into this working branch and will test building a genesis using the stuff in the `mainnet` repo to see if it fixes #15

the-frey:
Okay, the format of the community pool needs changing, but it's nearly there....

the-frey:
Ok, I think this is actually ready for review now. My only other concern is that I've noticed every account has its `account_number` set to 0 in the genesis, but they're incremented when generated elsewhere. Should we be enforcing uniqueness? I don't fully grok the purpose of that field, hence asking.

faddat (commented):
This looks good but will need to test it, as well.  

I think this would be a very useful module in the sdk, too.

dimiandre (approved):
Lgtm

PR #20, opened 2021-06-26
Create Dockerfile

faddat:


the-frey:
As I commented on discord:

Looks good, but given that it's specific to only the lucina testnet, I think it might belong better in the testnets repo - as in "if you want to run this testnet using docker, grab this dockerfile [link] and run" 

So I think I'd add it there & add a line to instructions. 'Cos we're likely to have at least one more testnet and it's easier to keep track of/update there. The production docker inspired by other projects in this main repo is a different thing and probably best to not get them confused, I would say.

This is really good work though, so we should get it up where people can use/see it! Just suggest testnets is the correct place.

faddat:
So I have long felt that repositories like Gaia are sorely lacking a very fast way to launch the node software.

I don't really think that this belongs in Chestnut but instead here, as the fastest way to run a node.

It is not suitable for use by validators.

faddat:
moving to faddat/instachainz

PR #21, opened 2021-06-26
Create instajuno.yml

faddat:


the-frey:
Same as for #20 I think these belong in the testnets repo (and might be best off in the same PR as the one workflow depends on the other)

faddat:
You are right that they should be in the same pull request and I think that you're wrong that they should go in the testnets repository.

When we talked about this, you made mention of these being used for validators but they're absolutely not for use  by validators.

Instead, this is to be the fastest path to having a node running. Think of it as an instant RPC endpoint for Juno.

faddat:
What would you say to me combining these into a single pull request because indeed it wasn't really a great style?

Again this is not something that I think that validator should be using It would be pretty much terrible for a validator to use these, and instead it's the fastest way to get a node.

the-frey:
Covered elsewhere by @faddat's work on https://github.com/faddat/instachainz so closing for now.

faddat:
It's also been done better in akash's cosmos-omnibus.

PR #26, opened 2021-07-10
Allow IBC on smartcontract

giansalex:
Currently, i get this error:

image

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -368,7 +368,7 @@ func New(
 
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
-	supportedFeatures := "staking"
+	supportedFeatures := "staking,stargate"
 	app.wasmKeeper = wasm.NewKeeper(
 		appCodec,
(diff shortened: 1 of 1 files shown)


the-frey:
LGTM. Any objections @JakeHartnell / @dimiandre ?

giansalex:
is it also possible to enable this flag?
https://github.com/CosmosContracts/Juno/blob/c6dde2025e178bba0e41c46cb35dcef6312b7cc0/app/app.go#L101

JakeHartnell:
Oh, we may need to make a proposal to enable IBC via governance. Keep this ticket open.

Can probably change the `ProposalsEnabled` flag.

the-frey:
If I understand @JakeHartnell correctly, that second commit should be on a separate branch...

the-frey:
Just so I understand @JakeHartnell - why would we need governance to enable IBC? A number of Hack Juno challenges seem predicated on it, and Juno is being talked up as being IBC enabled from network launch. It seems that would also include IBC for SCs...?

PR #37, opened 2021-08-14
0.43.0

faddat:
Recommend that we merge this to main, triage bugs like:

Screen Shot 2021-08-14 at 8 39 54 AM

then go

Chain runs, and we can then use the v0.43.0 "known good chain" to finish stuff up with the mint.


faddat:
this seems to merge smoothly.  

Oh right the wierd... thing.  1 moment.

PR #39, opened 2021-08-20
Enable x/wasm proposals

giansalex:


faddat (approved):
I approve but I think we are holding off on wasm till a while after launch.  Perhaps this should go into a branch, `postlaunch-wasm` or so?

JakeHartnell:
> I approve but I think we are holding off on wasm till a while after launch. Perhaps this should go into a branch, `postlaunch-wasm` or so?

Unless CosmWasm 1.0 comes out in the next week, I believe this will be the case.

PR #42, opened 2021-08-24
Latest starport and pre-0.43.0

faddat:


dimiandre:
> I strongly think we should use 0.43.0

Do you think we should wait for starport to support it or change version manually?

faddat:
I have experience changing the version manually.

dimiandre:
Merged! Thanks a lot!

PR #51, opened 2021-09-14
add custom /txs endpoint for wallet providers

dimiandre:
This PR add a custom `/txs` post endpoint to Legacy REST API that was removed by Cosmos-SDK v0.44. Since a lot of wallets still rely on that for transaction broadcasting we prefer to keep it for now.

Multisig + sequence security vulnerability should not affect this change

Specs for the custom endpoint
```typescript
{
  "tx": StdTx,
  "mode": "block" | "sync" | "async",
   // Below fields are newly introduced optional arguments.
   // To convert legacy tx to new proto tx, we need each signer's sequence number. 
   // If this field is not given, we fetch the sequence from the chain state. Normal 
   // users can skip this field as well, but the users who are sending faster than block 
   // time need to provide this field.
  "sequences": ["1", "2"],
  "fee_granter": "",
  "timeout_height": "10"
}
```

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -32,6 +32,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/auth"
 	"github.com/cosmos/cosmos-sdk/x/auth/ante"
 	authrest "github.com/cosmos/cosmos-sdk/x/auth/client/rest"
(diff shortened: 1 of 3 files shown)


giansalex:
Now, keplr has `no-legacy-stdTx` feature.
https://github.com/chainapsis/keplr-extension/issues/163

PR #54, opened 2021-09-20
Bump spm to v0.1.5

freshfab:
The `EncodingConfig` struct in `spm v0.1.4` is not compatible with `cosmos-sdk v0.44.0`.

PR #57, opened 2021-09-29
Update cosmos-sdk 0.44.1

giansalex:
Contains bug fixes , reference [cosmos-sdk@v0.44.1](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.44.1)


faddat (approved):
if this is a non-breaking change, we want it fast because I think it addresses iavl performance.

the-frey:
AIUI this needs to be merged before #54 - correct?

Issue #17, opened 2021-06-22
5% minimum validator comission


faddat:
This leads to a healthier network that is more difficult to capture, so I think we should get this in before cutting mainnet.

blockcreators:
So it looks like the community poll has changed back to 2.5%
Perhaps it can start with this, and a governance proposal can be put up after launch.

faddat:
Kinda good with just going with 5%.

But let's see.  I won't get this in till some time next week anyhow.

faddat:
I need to either get osmo's minimum commission into the comsos-sdk or we use the osmo fork.

faddat:
You know I should try to squeeze this one in I think it will improve the quality of the network and won't be difficult #43

nullmames:
i am all for this. i noted there are already gentx submissions with 0% commissions

nullmames:
at the risk of an infinite loop
https://www.mintscan.io/juno/proposals/3

leha2148:
@faddat the vote was accepted, but what will happen to the validators who set the maximum commission less than 5%, they will not be able to raise it? maybe give them the opportunity to do it 1 time?

nullmames:
@leha2148 the minimum commission will be introduced with the upcoming moneta upgrade. this means that any validator with less than 5% commission before the upgrade will have 5% commission after the upgrade, automatically.

the-frey:
Right so to implement this we need:

- [ ] Param change
- [ ] Upgrade handler

Anything else? Could use clarifying here so that it can be implemented

giansalex:
This is done

PR #75, opened 2021-10-30
Add upgrade handler

the-frey:
Add an upgrade handler for moneta alpha. Somehow got missed before.
After this is merged, app should be tagged `v2.0.0-alpha.2` and instructions updated/uni testnet validators informed.


dimiandre (approved):
Looks good to me, not sure how to handle on mainnet. We need also an upgrade for min commission

the-frey:
The issue for the commission change is #17 I think

giansalex (approved):

JakeHartnell (approved):

PR #81, opened 2021-11-08
Validator minimum commission

the-frey:
Upgrade handler in the vein of osmo solution, ante fix ported from stargaze.

Attempt to close off #17 so that we can test and ship with moneta.
Needs tagging as `v2.0.0-beta` if/when approved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ cosmoshub3.json
 exchanges.json
 juno_out.json
 bin
+
+# emacs editor config
+\#*\#
+.\#*
--- a/app/ante.go
+++ b/app/ante.go
@@ -1,0 +1,97 @@
+package app
+
+import (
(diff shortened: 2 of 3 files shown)


giansalex:
I see it as a temporary solution

https://github.com/cosmos/cosmos-sdk/pull/10422#issuecomment-963014427

the-frey:
@giansalex yep, when that gets merged we can switch over. In the meantime, we need to get validators on 5% ASAP, so would be nice to get this in with moneta if poss 🙏

faddat (approved):
Approving for testnet usage.  I really am not in love with this solution and it feels a bit funny to me.  

That said, I do not think that it will explode things and I do think that it will get the job done.

giansalex (approved):

JakeHartnell (approved):

Issue #85, opened 2021-11-13
Set upload fees for contracts higher


JakeHartnell:
It should cost more to upload a contract. I believe there is already a parameter for this? Let's update it on the Uni testnet.

If there is not a parameter, we should add this feature. Transactions should be cheap, but uploading a contract should be expensive.

blockcreators:
Governance can change it later but what do you think the initial cost per byte should be?
And how much would the average contract be? (byte cost * average contract bytes)

To make a simple token via junomint, how much do we expect it to cost?

the-frey:
...which I guess in practical terms needs I need to write the checklist lol

JakeHartnell:
> ...which I guess in practical terms means I need to write the checklist lol

Appreciate all your help on this. :pray:

JakeHartnell:
> ...which I guess in practical terms means I need to write the checklist lol

Let's try to keep the list here in GitHub issues. Let's try to use this Milestone: https://github.com/CosmosContracts/juno/milestone/1

the-frey:
okay so there's also the upload size to consider too.

In wasm;

```go
// DefaultMaxWasmCodeSize limit max bytes read to prevent gzip bombs
DefaultMaxWasmCodeSize = 600 * 1024 * 2
```
and

    junod query params subspace wasm maxWasmCodeSize --chain-id uni

confirms we are using default

```
key: maxWasmCodeSize
subspace: wasm
value: '"1228800"'
```

JakeHartnell:
> While the various costs are specified https://github.com/CosmWasm/wasmd/blob/57517b0c333e23aedd2bbccf80579cf75081241b/x/wasm/keeper/gas_register.go#L13
> 
> With reasoning https://github.com/CosmWasm/cosmwasm/blob/v1.0.0-beta/docs/GAS.md

Good find! Looks like this is maybe what we want:
```
	// DefaultCompileCost is how much SDK gas is charged *per byte* for compiling WASM code.
	// Benchmarks and numbers were discussed in: https://github.com/CosmWasm/wasmd/pull/634#issuecomment-938056803
	DefaultCompileCost uint64 = 3
```

the-frey:
Yup, although it's hard coded AFAICT which would mean forking. AFAIUI changes there are consensus-breaking, hence why they aren't params. That said, it's lateish in the day and my caffiene level is drooping so may have that wrong.

faddat:
hate that caffeine drooping feeling

[...]

PR #90, opened 2021-11-19
Override cw default compile costs

the-frey:
Addresses #85 

I've now checked that the node runs at least but will run some more checks over the weekend.

The bash scripts I'm now using to bootstrap the app in dev are pretty useful so I've committed and documented them because tbh people have been asking on discord & it's on my to-do list to improve the dev section of the docs.

Probably could use a once-over for sanity-checking before I spam a bunch of contracts at it.

Once we're happy with code integrity here and want to merge, there's the question of what to actually change the value to. I assume we'll want to raise it a bit more than what I've done here and run some tests on uni? @JakeHartnell - thoughts?


JakeHartnell (approved):
Yolo.

JakeHartnell:
Build and Deploy is failing because of this dumb file: https://github.com/CosmosContracts/juno/blob/main/.github/workflows/pages.yml

I vote we remove it as we're not really using the Starport vue stuff. :shrug:

app/wasm_config.go
```diff
@@ -0,0 +1,27 @@
...
+	return wasmkeeper.WasmGasRegisterConfig{
+		InstanceCost:               wasmkeeper.DefaultInstanceCost,
+		CompileCost:                DefaultJunoCompileCost,
+		GasMultiplier:              wasmkeeper.DefaultGasMultiplier,
+		EventPerAttributeCost:      wasmkeeper.DefaultPerAttributeCost,
+		CustomEventCost:            wasmkeeper.DefaultPerCustomEventCost,
+		EventAttributeDataCost:     wasmkeeper.DefaultEventAttributeDataCost,
+		EventAttributeDataFreeTier: wasmkeeper.DefaultEventAttributeDataFreeTier,
+		ContractMessageDataCost:    wasmkeeper.DefaultContractMessageDataCost,
+	}
```
giansalex, line 22: maybe we don't need to modify this, if new parameters will be added to `WasmGasRegisterConfig`
```suggestion
	gasConfig := wasmkeeper.DefaultGasRegisterConfig()
	gasConfig.CompileCost = DefaultJunoCompileCost

	return gasConfig
```

app/wasm_config.go
```diff
@@ -0,0 +1,27 @@
+package app
+
+import (
+	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
+)
+
+const (
+	DefaultJunoCompileCost uint64 = 4
```
JakeHartnell, line 8: Can we YOLO make this number 10.

JakeHartnell: :joy_cat:

JakeHartnell, line 8:
```suggestion
	DefaultJunoCompileCost uint64 = 10
```

faddat:
Agree about YOLO make it ten.

expensive is **good**.

the-frey:
I think as mentioned on TG that if this were tagged correctly, I can use this as the target binary for the instructions for the `moneta-beta` upgrade handler, meaning it doesn't need its own upgrade handler. Otherwise would maybe need one as iirc these param changes are consensus-breaking.

the-frey:
@faddat / @giansalex happy for these changes to go in?

giansalex:
@the-frey yes, must be included in moneta-beta upgrade

dimiandre:
Looks good to me! 

Can these params be modified by governance?

dimiandre (approved):

JakeHartnell (approved):

the-frey:
@dimiandre I think we could make them modifiable by governance in future, but for the moment they have to be changed in our codebase.

PR #93, opened 2021-11-23
dev-env: setup docker-compose

albttx:
Hey :wave:,

First MR on this project,

this allows to quickly setup a development environment using docker / docker-compose.

## How to use

```bash
export JUNOD_SEED_ADDR=$(junod keys show <wallet-name> -a)  
docker-compose up -d
```

I'd like to be able to use this docker-compose in every smart-contracts project I'll work on.

that would be amazing if the docker image could be on a docker hub. so the docker-compose can use `image: ...`


giansalex:
docker image is available here https://github.com/CosmosContracts/juno/pkgs/container/juno
```yaml
   image: "ghcr.io/cosmoscontracts/juno"
```

albttx:
Hello @giansalex, thanks for your answer.

Maybe it's better to have a docker-compose inside of juno that's use `build: ` to be able to quickly hack juno blockchain.

just re-build and restart to see your change on the blockchain.

But for a smart contracts project, it would be better to use the docker image.

and provide in the documentation how to work with juno showing a simple dockerfile who use an pre-built image 

What do you think ?

giansalex:
@albttx ok, but there is #94

albttx:
Hey, i just updated my MR now that I understand better all problematics

This allow all possibilities

- We are still able to run `junod` blockchain in local environments using the scripts (see `./scripts`) without docker
- For production environment, we have just a simple `junod` docker image containing the junod binary
- For hacking juno blockchain, we have the Docker image in `./docker/juno-blockchain` with the root `docker-compose.yml` file
- For peoples working on smart-contracts, they'll have the `./docker/juno-blockchain` image to quickly run the environment.

ps:  I use the `/scripts/setup_junod.sh` for setup to avoid having multiple identical scripts

What do you think ?

Ask for review: @faddat @the-frey @dimiandre @JakeHartnell @giansalex

faddat (approved):
Approved. 

Thanks so much for thinking about developer ergonomics.

giansalex:
I think it's done at #94, this must be closed

PR #94, opened 2021-11-24
Add compose file and restructure docs

the-frey:
This:

- Adds a compose file, meaning the local docker build is cached, and exposes all relevant ports
- Changes the setup script so that if it runs inside of an already-built container, already executed parts are skipped
- Instructions are moved out of the readme and into GitBook (I've parked the relevant sections in a change request there until this is merged)
- A couple typos I intro'd to the readme fixed by their removal (lol)

Thanks to @albttx for making me consider the utility of a `docker-compose` file even for a single-service project.

This PR supercedes https://github.com/CosmosContracts/juno/pull/93

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3"
 services:
   node:
     build: .
-    command: ./setup_and_run.sh
+    command: ./setup_and_run.sh juno16g2rahf5846rxzp3fwlswy08fz8ccuwk03k57y
     ports:
       - 1317:1317 # rest
       - 26656:26656 # p2p
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,9 @@
+version: "3"
+services:
+  node:
+    build: .
+    command: ./setup_and_run.sh
+    ports:
+      - 1317:1317 # rest
+      - 26656:26656 # p2p
+      - 26657:26657 # rpc
--- a/docker/setup_and_run.sh
+++ b/docker/setup_and_run.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-./setup_junod.sh
+./setup_junod.sh "$@"
 ./run_junod.sh
--- a/docker/setup_junod.sh
+++ b/docker/setup_junod.sh
@@ -8,29 +8,39 @@ CHAIN_ID=${CHAIN_ID:-testing}
 MONIKER=${MONIKER:-node001}
 KEYRING="--keyring-backend test"
 
-junod init --chain-id "$CHAIN_ID" "$MONIKER"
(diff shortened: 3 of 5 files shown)


albttx:
Hello,

Thanks @the-frey  for this update but I'm not 100% ok with it. let me explain why.

- this structure make it work ONLY inside the [juno git repository](https://github.com/CosmosContracts/juno) 

this is a major problem since the current project architecture is multi-repository oriented, every Juno project will be separate git repositories.
 
for example, I'm currently working on smart contracts for juno, I'd like to have a `docker-compose.yml` file inside that smart-contract repository to run easily the blockchain and deploy my smart-contracts on it (without having to have the Juno repository on my machine).

albttx:
Yes we need a pre-built docker image.

the docker image would require the entrypoint and cmd to start the blockchain,

easier the better :+1:

the-frey:
We have a pre-built image that is designed to be used in production, but its commands can be overridden. This PR improves our existing bootstrap process for local/dev running while not breaking the Dockerfile by adding things that aren't relevant for a production build. As for pre-built images, we already have them - but it is fair that we ought to tag specific versions better. I have raised this in #98

albttx:
Ok, so if the docker image if for production use, I think we should separate in multiple docker images

- one containing only the junod binary
- one with the blockchain configuration (that could be a multi stage build docker image using junod image builder)

What do you think?

faddat (approved):
I'm cool with this, there's no problem with it, but serious, serious question:

Anyone use setups like this?

If so, why?

I used to but have found them basically unnecessary.

albttx:
Hello @faddat, my changes are #93 

- Docker based environment allows easier hacking on a project
- Faster setup, onboarding

Also, I'm new to blockchain programming and i like to be able to test my smart contracts on a localhost environment.

As I said, the goal of this (#93) is to be able to have a docker-compose to test smart contracts in their respective repositories. this current MR don't allow to do that !

the-frey:
@faddat yeah this is actively being used by me at the mo to develop with.
@albttx your use-case is served by #101 which will provide standalone tagged builds, i.e. `docker <pull|run> ghcr.io/cosmoscontracts/juno:<tag>`

As it stands your PR requires somebody to check out the repo and use the compose file anyway, so I don't really follow, since that is what _this PR addresses_.

This has all gotten a bit complex. I wanted to take the good points raised in 93 and get some changes in in such a way that I knew there was nothing being broken and would be compatible with current changes to the docs.

There's currently a lot in flight and we admittedly don't do enough to specify requirements outside the core team. That's probably my fault.

JakeHartnell (approved):

faddat:
If it's being used then we should get it in even if I don't understand how or why

Issue #89, opened 2021-11-19
Build and Deploy CI is broken


the-frey:
More as a note to self to fix this later, but we've got a broken window, and leaving it red will result in us missing stuff that actually _is_ important before long.

the-frey:
`pages.yml` is gone as of #90 and #96 / #97 should close this off

PR #96, opened 2021-11-25
Update build.yml

faddat:
#89


the-frey:
Is it intentional that we no longer build on multiple target architectures in this?

the-frey:
From TG: rationale is that ARM is currently not working & this will speed up the build. ARM/matrix builds can be restored as/when support for CW is sorted.

the-frey (approved):
LGTM

JakeHartnell (approved):

Issue #98, opened 2021-11-25
Better tagging of containers in registry


the-frey:
At the moment there is only `latest` AFAICT

https://github.com/CosmosContracts/juno/pkgs/container/juno

Would be nice if this was created in such a way that tags were applied that match releases, for dev purposes

JakeHartnell:
#107 is potentially related?

the-frey:
![C-658VsXoAo3ovC](https://user-images.githubusercontent.com/2385077/144013927-aedac0ee-9566-4e2d-9992-6ba6fad36a24.jpg)

the-frey:
What remains to be seen is how the github CI improvements react to tags. It _should_ tag and release when you tag, but I don't know if that means it expects the tag to be tagged and pushed from the dev machine _when it lands on GitHub_

the-frey:
Huh, looks like it still isn't quite detecting the tags right. FFS.

Will add this back to my to-do list.

the-frey:
Aha. Tagging only works for _pushing tags_ and releases/tags via UI is a separate event. PLUS

> Note: The release event is not triggered for draft releases.

Which is why current draft isn't picked up.

the-frey:
Release tagging can now be seen working on `main`. Might disable PR tagging at some point however.

PR #101, opened 2021-11-26
Improve tagging

the-frey:
Addresses #98 

This should improve tagging to use the git tags. For some reason it's not triggering the PR CI though. Hate testing GH actions.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -39,6 +39,6 @@ jobs:
         platforms: linux/amd64
         push: true
         tags: |
-          ${{ steps.prepare.outputs.tag }}
-          ${{ steps.formatowner.outputs.lowercase }}/juno:latest
+          ghcr.io/${{ steps.prepare.outputs.tag }}
+          ghcr.io/${{ steps.formatowner.outputs.lowercase }}/juno:latest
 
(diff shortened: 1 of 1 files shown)


the-frey:
Hm, actually. Not sold on the action/plugin, will try something else.

the-frey:
Ready to review, think this one is going to require another iteration if something borks on main. Kind of a PITA

the-frey:
Cool, PRs are working (should mean it's easier/quicker to check the app integrity when we have changes to check on a small PR) and hopefully tags will be picked up when it's on main 👍

.github/workflows/docker.yml
```diff
@@ -1,40 +1,45 @@
 name: ci
 
 on:
+  pull_request:
```
giansalex, line 4: I'm not sure if docker image should published by `pr`, generally is when new release is created or push to main branch
Example: [Starport](https://github.com/tendermint/starport/blob/develop/.github/workflows/release-docker.yml)

the-frey: My reasoning was that quite often when I'm reviewing a PR I want to really quickly run it and see if it works. I guess it's only slightly quicker than checking out the branch though, so happy to remove - your call dude

JakeHartnell (approved):

faddat (approved):
I am utterly incapable of making this kind of GitHub action work right and I am approving it on principle

PR #103, opened 2021-11-28
Fix min-commission upgrade [moneta-beta]

giansalex:
Skip validations:
- max change rate:
- new rate cannot be changed more than once within 24 hours

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -718,6 +718,10 @@ func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
 		minCommissionRate := sdk.NewDecWithPrec(5, 2)
 		for _, v := range validators {
 			if v.Commission.Rate.LT(minCommissionRate) {
+				if v.Commission.MaxRate.LT(minCommissionRate) {
+					v.Commission.MaxRate = minCommissionRate
+				}
+
 				v.Commission.Rate = minCommissionRate
 				v.Commission.UpdateTime = ctx.BlockHeader().Time
 
--- a/app/app.go
+++ b/app/app.go
@@ -718,12 +718,8 @@ func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
 		minCommissionRate := sdk.NewDecWithPrec(5, 2)
 		for _, v := range validators {
 			if v.Commission.Rate.LT(minCommissionRate) {
-				comm, err := app.StakingKeeper.UpdateValidatorCommission(
-					ctx, v, minCommissionRate)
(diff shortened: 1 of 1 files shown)


blockpane (approved):

app/app.go
```diff
@@ -718,12 +718,12 @@ func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
 		minCommissionRate := sdk.NewDecWithPrec(5, 2)
 		for _, v := range validators {
 			if v.Commission.Rate.LT(minCommissionRate) {
-				comm, err := app.StakingKeeper.UpdateValidatorCommission(
-					ctx, v, minCommissionRate)
-				if err != nil {
-					panic(err)
+				if v.Commission.MaxRate.LT(minCommissionRate) {
+					v.Commission.MaxRate = minCommissionRate
```
zlounes, line 722: here you change the MaxRate, but you have to reset the value after have been able to change the commission rate

zlounes (changes requested):

app/app.go
JakeHartnell: I don't think it's a problem, because here he's checking if the max rate is _less than_ the min commission.

blockpane (approved):

app/app.go
```diff
@@ -718,12 +718,12 @@ func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
...
 				}
-				v.Commission = comm
```
zlounes, line 726: Where do you change the min commission rate?

zlounes (changes requested):

JakeHartnell (approved):

blockpane (approved):

app/app.go
JakeHartnell: Line 125.

app/app.go
giansalex: max-commission should not be less than the new commission rate.
just be consistent.

app/app.go
zlounes: Rate === MinRate?! OK then

blockpane (approved):

app/app.go
zlounes: But is not related to the Rate or MinRate, but MaxRate.

the-frey (approved):

app/app.go
zlounes: Oh yes ofc giansalex. Sorry!

zlounes (approved):

PR #104, opened 2021-11-28
Allow commission rates to exceed max and max change rates [moneta-beta] 

chillyvee:
This patch seems to start junod, but not sure if it's the right fix.

I also just saw the following which might be a better / more complete fix.
https://github.com/CosmosContracts/juno/pull/103


giansalex:
an error avoid to update commission

chillyvee:
> an error avoid to update commission

Agreed, this patch is no good since UpdateValidatorCommission fails first.  Your patch is better :)

PR #105, opened 2021-11-29
Roll moneta-alpha and moneta-beta into a single upgrade handler

the-frey:
As `moneta-alpha` required a no-op handler, I _think_ this might be all that is needed to merge `-alpha` and `-beta` into a single handler. We should probably not merge this yet though until we're 100% on how we should test it. I'm sort of erring towards relaunching `uni` at the moment.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -705,11 +705,7 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 
 // RegisterUpgradeHandlers returns upgrade handlers
 func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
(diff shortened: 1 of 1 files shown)


giansalex:
u also need to update here 
https://github.com/CosmosContracts/juno/blob/6d8f7f5354b4b7924d3ba162a5c70cc6e835c682/app/app.go#L554

the-frey:
D'oh, yeah, you're right. Serves me right for quickly grepping through the file in between meetings.

faddat (approved):
lfg

giansalex (approved):

PR #115, opened 2021-12-06
Moneta Upgrade Review PR

JakeHartnell:
This PR is to make it easier to visualize and discuss the changes made from `v1.0.0` to Moneta. It is not meant to be merged, but it is meant to have as many eyes on it as possible.


Dockerfile
```diff
@@ -15,14 +15,14 @@ WORKDIR /code
 COPY . /code/
 
 # See https://github.com/CosmWasm/wasmvm/releases
-ADD https://github.com/CosmWasm/wasmvm/releases/download/v0.14.0/libwasmvm_muslc.a /lib/libwasmvm_muslc.a
-RUN sha256sum /lib/libwasmvm_muslc.a | grep 220b85158d1ae72008f099a7ddafe27f6374518816dd5873fd8be272c5418026
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.0.0-beta/libwasmvm_muslc.a /lib/libwasmvm_muslc.a
```
giansalex, line 18: will need to update to `v1.0.0-beta2`

the-frey: will do it #122 which is the final gas tweak for mainnet I think

SECURITY.md
```diff
@@ -0,0 +1,32 @@
...
+
+If it is something that can be handled by a param change, discuss it on Discord in the `#governance` channel, and consider a governance proposal.
+
+## Standard priority bug 🐛
+
+For a bug that is non-sensitive and/or operational in nature rather than a critical vulnerability, please add it as a GitHub issue.
+
+If it is not triaged in a couple of days, feel free to tag `@the-frey` or `@jakehartnell`.
+
+## Critical bug or security issue 💥
```
elsehow, line 23: I think we could do with a more professional disclosure policy. Here's a nice example from Tendermint: https://tendermint.com/security/

It's *really* important to set specific expectations around disclosure timeline. Otherwise feelings get hurt and bugs get disclosed in an uncoordinated way.

app/wasm_config.go
```diff
@@ -0,0 +1,25 @@
+package app
+
+import (
+	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
+)
+
+const (
+	// DefaultJunoInstanceCost is initially set the same as in wasmd
+	DefaultJunoInstanceCost uint64 = 60_000
+	// DefaultJunoCompileCost set to a large number for testing
```
spacepotahto, line 10: Is this comment still accurate?

SECURITY.md
JakeHartnell: Feel free to make a PR.

app/wasm_config.go
JakeHartnell: Not for testing, it's just set to a large number initially until we feel good about everything.

SECURITY.md
elsehow: Done - #124

PR #118, opened 2021-12-10
update imports

faddat:


the-frey (approved):
Can't comment too heavily on the wasmd side of things other than to say that if we're happy with them, then from a Juno POV this bump will work 👍

JakeHartnell (approved):

PR #117, opened 2021-12-09
10x compile cost again

the-frey:
Will drop a note in TG about this.

PR #122, opened 2021-12-13
Change compile cost again

the-frey:
WIP - currently testing locally but want the docker image created by the PR

### DIFF
--- a/app/wasm_config.go
+++ b/app/wasm_config.go
@@ -8,7 +8,7 @@ const (
 	// DefaultJunoInstanceCost is initially set the same as in wasmd
 	DefaultJunoInstanceCost uint64 = 60_000
 	// DefaultJunoCompileCost set to a large number for testing
-	DefaultJunoCompileCost uint64 = 1000
+	DefaultJunoCompileCost uint64 = 100
 )
 
 // JunoGasRegisterConfig is defaults plus a custom compile amount


the-frey:
Okay, see very long-winded hacking discussed on #117

the-frey:
So locally, with validator gas fees set to `0.025ujunox` 

    TXFLAG="--gas-prices 0.01$DENOM --gas auto --gas-adjustment 1.5 -y -b block --chain-id $CHAIN_ID --node $RPC"

FAILS

    insufficient fees; got: 476050ujunox required: 1190124ujunox: insufficient fee"

Upping fees:

     TXFLAG="--gas-prices 0.1$DENOM --gas auto --gas-adjustment 1.5 -y -b block --chain-id $CHAIN_ID --node $RPC"

Succeeds. The cost to the key for sending is:

'4760494`

the-frey:
TL;DR of options:
1. This ⚖️ 
2. Reduce down to wasmd default (much lower than `100`, which is a 10x of our original bump) 😓 
3. YOLO and keep the 10x version that this lowers (`1000`) with the assumption that we may need a param change on mainnet and might be opening ourselves up to an attack vector 🤷

faddat (approved):

giansalex:
so revert #117  and back to 100 ...

the-frey:
@giansalex yeah we can do that - we probably want the version bump in the docker container and we _definitely_ want the fees change (means this stuff can be locally tested) - but can put those in a separate PR?

JakeHartnell (approved):

PR #126, opened 2021-12-16
Bug: submessage failures

the-frey:
Currently investigating submessage failures in larger CW contracts

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	filippo.io/edwards25519 v1.0.0-beta.2 // indirect
 	github.com/99designs/keyring v1.1.6 // indirect
 	github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d // indirect
-	github.com/CosmWasm/wasmvm v1.0.0-soon2.0.20211209162446-b96c92d21aaf // indirect
+	github.com/CosmWasm/wasmvm v1.0.0-beta2 // indirect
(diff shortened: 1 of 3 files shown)


the-frey:
Clearly we need to get this fixed on mainnet. This could be a reasonable temporary solution, but of course those more knowledgeable feel free to weigh in.

cc @giansalex @faddat

the-frey:
also cc @JakeHartnell

go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	filippo.io/edwards25519 v1.0.0-beta.2 // indirect
 	github.com/99designs/keyring v1.1.6 // indirect
 	github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d // indirect
-	github.com/CosmWasm/wasmvm v1.0.0-soon2.0.20211209162446-b96c92d21aaf // indirect
```
JakeHartnell, line 30: This is the problem. @faddat, can upgrade the Notional Labs fork to use `v1.0.0-beta3`? It's the latest release. https://github.com/CosmWasm/wasmvm/releases

7 is a lucky number.

faddat: so, we're using beta2 currently, but golang tags stuff kinda odd.

Here's the pr for beta3:
[
](https://github.com/CosmosContracts/juno/pull/127)

JakeHartnell:
Confirmed working without a downgrade. @giansalex found the issue.

PR #128, opened 2021-12-16
wasmvm beta 3

faddat:


go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/CosmWasm/wasmvm v1.0.0-soon2.0.20211209162446-b96c92d21aaf // indirect
```
giansalex, line 30: should be `github.com/CosmWasm/wasmvm v1.0.0-beta3` ?

faddat: nope, it is go doing its tag breaking thing.  but let me check the commit height to be certain.

faddat: ah.  

More fixing needed.

faddat: oop.  That would have broken everything...

faddat:
night everyone.  Never actually seen it require an explicit update for an indirect dependency.

JakeHartnell (approved):

JakeHartnell (changes requested):
Ok, this is sadly not building...

PR #129, opened 2021-12-16
Pass in MsgServiceRouter from BaseApp

JakeHartnell:
Maybe this?


faddat:
definitely maybe

faddat (commented):
well-worth testing in combination with #128

JakeHartnell:
Doesn't fix issue...

PR #133, opened 2021-12-18
Use latest Juno wasmd

JakeHartnell:


faddat (commented):
going to make a PR to this PR that adds rocksdb support.

faddat (approved):
Looks good.  Thanks for the patch!

Issue #132, opened 2021-12-17
Contract Address Length


the-frey:
It is shorter since `v2.0.3` -> we need to increase in order to be forwards compatible with mainline.

Terse ticket for now. Let's discuss fix and upgrade path.

cc @dimiandre @JakeHartnell @faddat @giansalex

the-frey:
See https://github.com/cosmwasm/wasmd/issues/685#issuecomment-981547822

giansalex:
v1.0.0-juno | v1.0.0-juno3
![image](https://user-images.githubusercontent.com/14926587/146558183-63978014-a3e5-4129-b82b-98a5fe7ad31a.png)

https://github.com/notional-labs/wasmd/blob/v1.0.0-juno3/x/wasm/keeper/keeper.go#L880

the-frey:
So in terms of an upgrade path - we need to write a migration that:

- Looks up all contracts currently instantiated
- Replaces their contract addr with a 32 byte version?

Luckily not many contracts deployed yet 🤔

the-frey:
fixed in wasmd and upgraded in #135

PR #135, opened 2021-12-19
Add moneta-patch base upgrade handler

the-frey:
A basic upgrade handler for `moneta-patch` - if there are migrations required to e.g. try and change existing contract addresses then we will need to change this.

Second question - we can remove the `moneta` upgrade handler now, right?

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -734,6 +734,10 @@ func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
 		}
 		return app.mm.RunMigrations(ctx, cfg, vm)
 	})
+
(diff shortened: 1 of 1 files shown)


the-frey:
I assume if we need to update addresses lengths it might be hairy, cos we'd be updating them in place in storage based on instance id, a la https://github.com/CosmosContracts/juno/issues/132

But also unclear if that is also a hard requirement. Feels like we can't both have contracts that are 20 and 32 bytes without borkage, but I am by no means an expert.

JakeHartnell (approved):

giansalex (approved):

PR #137, opened 2021-12-23
Security Patch

JakeHartnell:


the-frey (approved):

dimiandre (approved):
Tested locally and upgrade work as expected. 

It's a go for me, let's upgrade with this. I'll continue testing during the day

PR #142, opened 2022-01-19
Maintain 1.0

faddat:
This is a shot at maintaining Juno 1.0, which will be necessary for archive nodes and sync processes for years to come. It's nothing more than version bumps and the additon of rocksdb via cosmos﻿


faddat:
I think this fails because it doesn't have wasm, but someone should check it out and merge it anyhow.

JakeHartnell:
Are we sure this won't cause issues with the version bump? Any chance this might be consensus breaking? I mean we would still have the older version tagged... so perhaps not much of an issue.

faddat:
I've synced it end to end myself

JakeHartnell (approved):

PR #150, opened 2022-01-28
Add validation to min-commission antehandle

giansalex:


the-frey (approved):
Looks reasonable. I say we test it on uni with the other bits and pieces

JakeHartnell (approved):

PR #151, opened 2022-01-29
Update app.go

faddat:
NB:  I might've misunderstood this:

https://github.com/cosmos/cosmos-sdk/commit/222fd6cd358b05af8ba08deb64f7ee311c91755f#commitcomment-65011785

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -750,10 +750,31 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
 	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
 }
 
-// RegisterUpgradeHandlers returns upgrade handlers
-func (app *App) RegisterUpgradeHandlers(cfg module.Configurator) {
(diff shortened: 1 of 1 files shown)


giansalex:
From [v0.45 docs](https://github.com/cosmos/cosmos-sdk/blob/v0.45.0/docs/core/upgrade.md)

> For a new module `foo`, `InitGenesis` is called by `RunMigration` only when `foo` is registered in the module manager but it's not set in the `fromVM`. Therefore, **if you want to skip `InitGenesis` when a new module is added to the app, then you should set its module version in `fromVM` to the module consensus version:**
> ```go
> app.UpgradeKeeper.SetUpgradeHandler("my-plan", func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
>     // ...
> 
>     // Set foo's version to the latest ConsensusVersion in the VersionMap.
>     // This will skip running InitGenesis on Foo
>     fromVM[foo.ModuleName] = foo.AppModule{}.ConsensusVersion()
> 
>     return app.mm.RunMigrations(ctx, fromVM)
> })
> ```

But there are no new modules, and I don't see new any consensus version in SDK modules, and no migrations scripts
[v0.44.5...v0.45.0](https://github.com/cosmos/cosmos-sdk/compare/v0.44.5...v0.45.0)

faddat:
Thank you for that @giansalex.

PR #152, opened 2022-01-31
Tidy deps

the-frey:
Even though it works on a laptop and in CI, some stale state issues on a linux box that required 

    go mod download github.com/rs/zerolog

Even though this is present. So figured tidy would be a good idea.

Worth noting, this builds fine in:

- CI
- Local laptop
- Debian box

So suspect it's just long-lived boxes that have the issue 🤷 
Even so worth tidying up pre next release.


giansalex (approved):

giansalex:
we should include the moneta upgrade-handler https://github.com/CosmosContracts/juno/commit/e6b8c212b178cf575035065b78309aed547b1335, to update validator commission.

the-frey:
@giansalex you mean to ensure that if people snuck in between moneta and this upgrade their min commission is updated?

giansalex:
@the-frey yes, if someone does it on mainnet before the next upgrade.

the-frey:
@giansalex #153

JakeHartnell (approved):

faddat (approved):
gotta be tidy!

PR #153, opened 2022-02-01
Add forced commission update to lupercalia

the-frey:
As pointed out by @giansalex it was possible to statesync past the moneta handler.
This catches validators that might have tried to dodge the handler - although it should be noted that it looks like none have tried to do so/been able to do so on mainnet.


faddat (approved):
Awesome and great catch to everyone who flagged this.

the-frey:
@dimiandre / @giansalex okay with you?

giansalex (approved):

PR #147, opened 2022-01-27
Remove custom /txs endpoint

giansalex:
This reverts pull request #51

- Keplr supports cosmos-sdk 0.44x chains
- The cosmoshub mainnet (sdk v0.44) don't have this endpoint

PR #155, opened 2022-02-02
Revert "Remove custom /txs endpoint"

the-frey:
I don't think it's what crashed the chain but Keplr is no longer working for sends on my local machine and think this is the culprit. Maybe?

Reverts CosmosContracts/juno#147

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -16,6 +16,7 @@ import (
 	tmos "github.com/tendermint/tendermint/libs/os"
 	dbm "github.com/tendermint/tm-db"
 
+	customAuthRest "github.com/CosmosContracts/juno/custom/auth/client/rest"
 	"github.com/CosmosContracts/juno/docs"
 	"github.com/CosmosContracts/juno/x/mint"
 	mintkeeper "github.com/CosmosContracts/juno/x/mint/keeper"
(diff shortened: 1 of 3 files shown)


giansalex:
@the-frey did you add the "no-legacy-stdTx" feature?

giansalex:
works with [these features](https://github.com/giansalex/juno-cw20/blob/24cee0cb316cd54f65eac7c47bc272a771f71cd3/src/main.js#L228)

the-frey:
Right, but should we be telling all app devs to update? Seems poor form to just break their stuff. Might be better to work out a smoother upgrade path. Thoughts @JakeHartnell ?

giansalex:
since Juno is natively on keplr, this will not happen on mainnet.
I don't think this should be reverted just for testnet, we can add docs for this, and share the correct configuration for suggest chain in keplr

the-frey:
Yeah, could be persuaded by that approach. A bit annoying for app devs, but not the end of the world. Let's see what Jake's take is...

the-frey:
@JakeHartnell / @dimiandre what do you think for this? It looks like app devs will have to add a bit of extra config for keplr to work.

the-frey:
Ah yeah I guess we're on 45+ anyway

the-frey:
@giansalex getting reports that this is broken on local and `no-legacy-stdTx` feature does not fix it. Do you know of another workaround? Otherwise going to re-open and move to revert this.

giansalex:
first we should know what the error is.  
maybe it is related to the wasm REST change, and reverting this will not fix it.
i will ask in discord for more details

the-frey:
I'm now beginning to think it's actually that the REST server isn't enabled. I'm going to add it to the dev config for Juno as a flag, and then see if turning that on solves it 🤔

giansalex:
ok, anyway I have asked in discord, if they give more details.

the-frey:
I've put a PR in that enables the REST server if you set the CORS flag for local dev. That seems to fix the problem for local chains, so I reckon that's the quickest fix for local development.

PR #156, opened 2022-02-08
Bump wasmvm to 1.0.0-beta6

the-frey:
This should mean we can now test the chain upgrade on `uni-2`, hopefully without a chain halt 🙃


giansalex (approved):

go.sum
```diff
@@ -84,6 +84,8 @@ github.com/CosmWasm/wasmd v0.23.0 h1:yOksa/TORt2BtXjUKKqcJ67d6O+Jv2Y6wT1Ekm1Como
 github.com/CosmWasm/wasmd v0.23.0/go.mod h1:kNDnMAQDJyVek9k6SxCNijMCzOROzzUGBRT8r/vr7oY=
 github.com/CosmWasm/wasmvm v1.0.0-beta5 h1:38M8z89LB5cFMYB5vfjewMzz9Pr8TB1QBHdjnrWnkas=
 github.com/CosmWasm/wasmvm v1.0.0-beta5/go.mod h1:mtwKxbmsko1zdwpaKiRkRwxijMmIAtnLaX5/UT2nPFk=
+github.com/CosmWasm/wasmvm v1.0.0-beta6 h1:JYh9Z7qF6h9a2sOOx9hZRwCVS4aFdd+aiwrIXznLn8k=
+github.com/CosmWasm/wasmvm v1.0.0-beta6/go.mod h1:mtwKxbmsko1zdwpaKiRkRwxijMmIAtnLaX5/UT2nPFk=
```
webmaster128, line 88: Do the wasmvm v1.0.0-beta5 entries disappear with `go mod tidy`?

webmaster128 (approved):
LGTM

go.sum
the-frey: Good point.

webmaster128: Beautiful. Good to see only one version is in use.

the-frey: @webmaster128 I think your review is now stale, sorry about that - if you're happy to add the tick again 😂 ...

webmaster128: I'm grey because I'm a gues with no permissions here. It was gray all the time.

the-frey: Ah, so that's what that means! TIL

JakeHartnell (approved):

PR #157, opened 2022-03-01
new sdk, tm, ibc-go

faddat:
This PR updates:

* Cosmos-sdk
* tendermint
* ibc-go
* tm-db

Because it updates tm-db, it also means that we can clean up go.mod and remove the gorocksdb replace.

I recommend combining it with #154 before any upgrade to the Juno mainnet.


the-frey:
@faddat what's the dependency order? Does 154 need merging first? Looked at that and it seems good to go.

PR #158, opened 2022-03-10
Upgrade mainline wasmd version to 0.24.0

the-frey:
Move to new version and clean up some of our dep bumps now that mainline is on beta7 and current SDK.


JakeHartnell (approved):

faddat (approved):
lgtm

PR #166, opened 2022-03-19
Interchain Accounts

faddat:
adds interchain accounts to lupercalia upgrade.


faddat:
this work needed to be started fresh because I can't seem to find it.

PR #167, opened 2022-03-21
Interchain Accounts / lupercalia

faddat:
This PR will bring ICA to Juno, as well as:

* wasmd v0.24.0
* ibc-go
* an upgrade handler folder to better organize code
* go 1.18
* tighter linting

This PR should be applied after a governance "Software Upgrade" proposal to juno passes.  

The governance "Software upgrade" proposal should not be made until this code has been merged to Juno's main branch.


faddat:
Awesome @nghuyenthevinh2000!

Ok, now we just add the ica stuff which is here:

https://github.com/cosmos/ibc-go/blob/main/testing/simapp/app.go

faddat:
we should also use wasmd from the same commit (nearly) as Osmosis:

faddat:
I am making a cosmos-sdk fork with @pomvn's fast iavl patch only, so we don't depend on osmosis-sdk

faddat:
https://github.com/notional-labs/cosmos-sdk/tree/v0.45.1-junofast

giansalex:
we can keep this pull request only for ibc-go v3 and ICA, and move iavl and other forks to other pull requests.
I also heard that iavl fork throws app_hash errors in osmosis, or does it have another cause?

go.mod
```diff
@@ -123,7 +124,9 @@ require (
 )
 
 replace (
-	github.com/cosmos/iavl => github.com/cosmos/iavl v0.17.3
+	github.com/cosmos/cosmos-sdk => github.com/notional-labs/cosmos-sdk v0.45.1-allfast
+	github.com/cosmos/iavl => github.com/osmosis-labs/iavl v0.17.3-osmo-v5.0.20220321022426-8371754f3299
+	github.com/cosmwasm/wasmd => github.com/notional-labs/wasmd v0.21.1-0.20220319082325-607293cd0bfe
```
the-frey, line 129: Why are we moving off mainline wasmd?

The whole point of lupercalia was to move _to_ mainline wasmd and collect security patches along the way.

Just trying to understand reasoning here :)

giansalex: these forks should be removed from this pull request

faddat: ah ok so the reasoning is interchain accounts

no fork no ica

wdyt

faddat:
no one really knows about those apphash errors, we can take fast iavl out

nguyenvuong1122000:
@faddat, i think we should remove iavl from wasmd

faddat:
yes you're right and thanks very much vuong. 

It was kind of all or nothing with fast iavl.

faddat:
@vuong177 @nghuyenthevinh2000 

Hey guys, this needs to be reworked.  I'll click around a little, and see how much is needed.

faddat:
Hey, I am going to close this in favor of #200 

While writing that, it became apparent that it might as well be this.

PR #168, opened 2022-03-22
Add lupercalia beta handler to allow for testnet upgrade

the-frey:
This _should not be merged_. 

It is set as a draft PR purely so:

- changes are clear
- docker container will be built
- very obvious what the git tag refers to

This handler allows us to run a second upgrade on `uni-2` to test the additional things we want to sneak into lupercalia.


joshlopes:
Just a suggestion but for `This should not be merged.` you should add `[WIP]` to the PR title so people know that is not supposed to be merged. You also have automated gh actions that prevent that: https://github.com/apps/wip

PR #174, opened 2022-03-26
Proposal 17 Implementation

joeabbey:
Handles:
 - [x] Static Delegations
 - [x] Unbondings
 - [x] Redelegations
 - [x] Reclamation of Rewards

Tested on `scottie-1` testnet.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -16,6 +16,7 @@ import (
 	tmos "github.com/tendermint/tendermint/libs/os"
 	dbm "github.com/tendermint/tm-db"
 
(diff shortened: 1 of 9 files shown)


the-frey:
Just a note on this - code looks correct.

The name in the handler will likely need to change.

We are using `lupercalia` for Lupercalia Phase 1.

Assuming the Unity Prop is the one that gets tabled first, the handler for Phase 2 will be called `unity`.

If Unity Prop is rejected, it will likely be `some-other-name`.

joeabbey:
Closing in favor of #186

PR #175, opened 2022-03-29
Update CI script to also perform executes

the-frey:
References a newer version of the target project that tests more things:

- store
- instantiate
- execute


faddat (approved):
looks great

PR #177, opened 2022-04-04
Attempt to fix local REST server issues

the-frey:
Some users have reported Keplr issues even with `no-legacy-stdTx` set, and our team has encountered the same. It looks like it's simply a case of needing the REST server enabled, so makes sense to bundle it in with CORS config.

This PR is a placeholder so a container gets built that I can test locally.

### DIFF
--- a/docker/setup_junod.sh
+++ b/docker/setup_junod.sh
@@ -30,9 +30,14 @@ else
 fi
 
 APP_TOML_CONFIG="$HOME"/.juno/config/app.toml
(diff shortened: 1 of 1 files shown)


the-frey:
Update: tested it locally and this provides a fix for the local issues. Turns out the REST server wasn't running. Which... you would expect to be running if you set CORS config.

So this solves the issue in terms of what users expect and what we say you need to set in the docs for local dev.

giansalex (approved):

faddat (approved):
👍

PR #179, opened 2022-04-06
Reinstate dockerfile

the-frey:
I think we probably jumped the gun on ripping this all out and want to restore it ASAP.

Will probably need a few tries to get everything working again.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,51 @@
+name: ci
+
+on:
+  pull_request:
+  release:
+    types: [published, created, edited]
+  push:
+    tags:
+    branches:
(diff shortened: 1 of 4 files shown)


JakeHartnell (approved):
LGTM

the-frey:
This supercedes #175

JakeHartnell (approved):

the-frey:
Will want to get this merged off the bat but observe that this means we can now do ARM builds and that I should document how to do so when our hair is not on fire (pass a flag to compose/pass the build-arg to the main container).

dimiandre (approved):
LFG

giansalex (approved):

PR #182, opened 2022-04-12
Spike tidy CI action

the-frey:
It was recommended we add a tidy action to check deps. This should work, but might need a future iteration.


JakeHartnell (approved):
Very nice.

faddat (approved):
makes a ton of sense on a dozen levels

PR #188, opened 2022-04-22
Bump to wasmd version for unity

the-frey:
Updates our wasmd version to the pre-release `wasmd@v0.27.0-junity.0` tag which has the fixes we need.

A note on the naming:

> unity &gt; rc, so [we] need to go up in the alphabet to J(uno) Unity

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/CosmosContracts/juno
 go 1.17
 
 require (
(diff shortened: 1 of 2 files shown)


dimiandre (approved):

webmaster128 (approved):

faddat (approved):
sudo!

PR #200, opened 2022-05-14
v6: ibc-go v3, latest cw

faddat:
This PR:

* removes every trace of cosmoscmd
* is a state breaking upgrade to juno v6
* adopts a cmd folder layout in the style of wasmd itself


faddat:
Breaking into subcomponents.

PR #201, opened 2022-05-17
Bump version of wasmd to 27.rc3 and update antehandler

the-frey:
Need this in so we can test on testnet ahead of final version.


app/app.go
```diff
@@ -386,11 +386,18 @@ func New(
...
 	// Create Transfer Keepers
 	app.TransferKeeper = ibctransferkeeper.NewKeeper(
-		appCodec, keys[ibctransfertypes.StoreKey], app.GetSubspace(ibctransfertypes.ModuleName),
-		app.IBCKeeper.ChannelKeeper, &app.IBCKeeper.PortKeeper,
-		app.AccountKeeper, app.BankKeeper, scopedTransferKeeper,
+		appCodec,
+		keys[ibctransfertypes.StoreKey],
+		app.GetSubspace(ibctransfertypes.ModuleName),
+		app.IBCKeeper.ChannelKeeper,
+		app.IBCKeeper.ChannelKeeper,
```
joeabbey, line 393: Do we need two?

joeabbey: Yeah, just saw: https://github.com/cosmos/ibc-go/blob/v3.0.0/modules/apps/transfer/keeper/keeper.go#L35

joeabbey (approved):
LGTM, (assuming we're not activating ICA)

dimiandre (approved):
Lgtm!

PR #206, opened 2022-05-18
✅ use go module version paths

faddat:
This PR uses standard go module versioning so that others don't accidentally use the wrong version of juno when building stuff on or related to juno.


.github/workflows/docker.yml
```diff
@@ -46,6 +46,6 @@ jobs:
         with:
           context: .
           platforms: linux/amd64
-          push: true
+          push: ${{ github.event_name != 'pull_request' }}
```
the-frey, line 49: Please remove this as it isn't relevant to this PR 👍

the-frey: Other than that, looks good.

faddat:
Force push context:

instead of updating <stuff> I just checked out main and applied the updates directly to it.  This should be ready to rock.

the-frey (approved):
LGTM

joeabbey (approved):
vNice

PR #209, opened 2022-05-19
Update docker.yml

faddat:
Updates the docker build process so that:

* tests pass
* we don't try to upload images except for the main branch.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -46,6 +46,7 @@ jobs:
         with:
           context: .
           platforms: linux/amd64
-          push: true
+          push: ${{ github.event_name == 'pull_request' }}
+
           tags: ${{ steps.meta.outputs.tags }}
           build-args: arch=x86_64


faddat:
This PR makes all of the other PR's not fail.

the-frey:
Sorry, but the solution to this is to make it so it doesn't run for external contributors.

Also you have access to the Juno repo so can just make PRs that _do_ pass.

The PR building is very intentional, and used for testing before merging.

the-frey:
Astonishingly there's no way in Github actions to mark an action step as not required like you can in Travis/Circle. Seeing if there's a workaround now...

the-frey:
Update 2 - targetting syntax for PRs only seems to work for the _target branch_ so it's pretty unclear to me if there is a good solution to this other than saying "core team should commit from within the Juno repo" until the two tracking issues for that & the other issue above are actually dealt with by github 🤷

giansalex:
> Sorry, but the solution to this is to make it so it doesn't run for external contributors.

```yaml
continue-on-error: true
```
could be used

faddat:
I can't make pr's that pass @the-frey 

far as I know, no one can

faddat:
What this does, is it makes the docker container build but not push.

faddat:
I'd like for the full CI suite to run for externals.  Just not **do** anything, like push the docker image.

the-frey:
@faddat what I meant was that some of us use the juno repo and push branches to it. Those work - and you can see they do. You have the ability to do that.

@giansalex I looked at that command but from what I can tell it skips the step but still marks the build as failed (why?) there's a long thread of angry people wanting that changed in the GH actions issues section. If that's changed, then cool - probably what we need.

.github/workflows/docker.yml
```diff
@@ -46,6 +43,6 @@ jobs:
         with:
           context: .
           platforms: linux/amd64
-          push: true
+          push: ${{ github.ref == 'refs/heads/master' }}
```
the-frey, line 46: We don't have a master branch...?

the-frey:
FWIW on approach 1 - it seems like we could make it triggerable and remove the PR element (I think) using the syntax froom https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

faddat:
There's a long thread of people angry about everything about actions and they should be <3

Github should switch to gitlab runners...

I guess my vibe here is:

that which doesn't push the image, cannot harm us or the users?

the-frey:
Okay I think for now the solution is to also add the manual run trigger I pointed to above, because then AFAIUI when that's in main we can target a PR branch to build, and then remove the push filter for main _and_ the PR build `on` trigger.

the-frey:
I will leave a note to self to rejig the contract CI harness as well with some additional things that we test elsewhere (migration, gov prop, etc)

PR #215, opened 2022-05-23
Integrate ICA changes

the-frey:
This one is going to need a bunch of testing.

At first blush, no regressions, though there's some caveats with auth impl, and I guess we won't know if it really works until we test it, huh?


faddat:
We probably need to implement intertx, like in wasmd.

I think that serves as the "controller auth" piece.

the-frey:
Cool, will look at that today 👍

faddat:
2nd nb:  intertx has a bunch of cigarette label esque warnings on it. 

That said I am ~95% sure that it's being integrated because afaik, you can have a host without it, but cannot do the controller bit.

@catShaark may be able to clarify to some degree.

the-frey:
Yeah, I guess my thinking here was:

- finish integrating
- work out some way of testing this on uni-3 and/or using strangelove's amazing IBC-a-tron
- change it if we find bugs

Sound reasonable?

Will have this PR done with intertx in a few mins. 🕐

the-frey:
I think it's likely there will be another version of intertx we will need to upgrade to... but hey, should be enough to test for now.

joeabbey (dismissed):
This PR will get the ICA functionality integrated, but I think we need an authmodule as well before we can release it.

https://github.com/cosmos/ibc-go/blob/main/docs/apps/interchain-accounts/auth-modules.md

The authentication module must:
```
Authenticate interchain account owners
Track the associated interchain account address for an owner
Claim the channel capability in OnChanOpenInit
Send packets on behalf of an owner (after authentication)
```

Reading through the [middleware](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/ibc_middleware.go) there's definitely callbacks to the app which I don't think are implemented in juno.

Perhaps we should keep this in the `intergrate_ica_changes` branch until we have the authmodule implemented.

If this is not a requirement for integration, I'm happy to Approve this PR.

the-frey:
Yeah, that's my understanding. Hence why I suspect it might:

a) fall over in the wild
b) have a series of changes as people find issues

So we will likely have to change this before primetime

joeabbey (commented):
Looking at https://github.com/osmosis-labs/osmosis/pull/1564/files

I think we also need an upgrade handler.

the-frey:
Ah, I think you're right. I guess we also need to add some wasm message types into that registration block. Will look at it in a sec.

the-frey:
Also @joeabbey there's also the question of whether we go with just the first half of this (i.e. Host implementation) first, so we don't have to worry about the auth part. But it does feel like Juno has many use cases for being the controller chain.

app/app.go
```diff
@@ -756,8 +824,49 @@ func (app *App) RegisterTendermintService(clientCtx client.Context) {
...
+				sdk.MsgTypeURL(&authz.MsgExec{}),
+				sdk.MsgTypeURL(&authz.MsgGrant{}),
+				sdk.MsgTypeURL(&authz.MsgRevoke{}),
+				// wasm msgs here
+				sdk.MsgTypeURL(&wasmtypes.MsgStoreCode{}),
+				sdk.MsgTypeURL(&wasmtypes.MsgInstantiateContract{}),
+				sdk.MsgTypeURL(&wasmtypes.MsgExecuteContract{}),
+				sdk.MsgTypeURL(&wasmtypes.MsgMigrateContract{}),
+				sdk.MsgTypeURL(&wasmtypes.MsgUpdateAdmin{}),
+				sdk.MsgTypeURL(&wasmtypes.MsgClearAdmin{}),
```
the-frey, line 857: I think we maybe only want `MsgExecuteContract` here actually

the-frey:
So - proposal here @joeabbey and @faddat is to park this branch for now and just integrate the ICA host portion (i.e. the very first couple of commits), _with_ the required registration handler for the native modules we want, _plus_ wasm execute only.

Then we have a scaffold for then investigating auth further for the controller part.

Thoughts?

app/app.go
joeabbey: Yeah this seems sufficient:
```
				sdk.MsgTypeURL(&wasmtypes.MsgStoreCode{}),
				sdk.MsgTypeURL(&wasmtypes.MsgInstantiateContract{}),
				sdk.MsgTypeURL(&wasmtypes.MsgExecuteContract{}),
```
Initially simply allow the storing, instantiation, and execution of contracts through ICA

faddat (commented):
Hey, I would like to approve this, and fully support ica, as soon as there's a little more clarity on best practices.  I have also resolved conflicts with main.

faddat:
oh dear.  I did something wrong.  I will fix your branch, sorry!

faddat:
All fixed.  @the-frey should we try and maybe consult with the ibc-go team or the wasmd team on this?

It has always felt fundamentally okay to me, but I understand the decision to go with the more conservative approach so am somewhat of mixed minds.

the-frey:
I think let's not rush it and look to integrate ICA host now & do ICA controller in v8 unless somebody tells me I'm being overly paranoid. Will ask the Confio guys for advice

faddat:
Works for me.

faddat:
this will be in #232 though still not 100% certain what the net effect will be

aidan-calculated:
@the-frey Can we look at getting this over the line again soon? Kujira ecosystem would love to integrate with Juno using ICA!

Reecepbcups:
On ICA v0.1.7, we can move to v0.2.5 for IBC v4 once we merge wasmd30+ IBCv4 + IBCFees PR #387

the-frey:
Thanks @Reecepbcups for bringing this up-to-date with the new juno structure. 

I think the outstanding questions we had on this were: 

- comfort level with running chain as controller
- comfort level with maturity of intertx (anybody running it in prod yet?)

Other than that I guess we need an upgrade handler to whitelist messages and then to update the command that adds those to genesis files for utility purposes 🤔

the-frey:
Personally, happy to run the chain as a controller, still _mildly_ concerned about intertx

Reecepbcups:
@the-frey Running as controller for Uni, 
The command things were already done in the `add-ica-config` yes? upgrade handler is also done

the-frey:
Yep, so the add ica config is basically called when genesis files need the current mainnet messages added. So er, after or at the time/tag that an upgrade handler adds those to mainnet then they should be additionally added to the CLI cmd

the-frey:
I guess the messages we want will all already be enabled in this case yes derp

JakeHartnell (approved):
Can't wait for this!

faddat (approved):

app/upgrades/v12/upgrades.go
```diff
@@ -46,6 +59,49 @@ func CreateV12UpgradeHandler(
...
+		// create ICS27 Controller submodule params, controller module not enabled.
+		controllerParams := icacontrollertypes.Params{
+			ControllerEnabled: true,
+		}
+
+		// create ICS27 Host submodule params
+		hostParams := icahosttypes.Params{
+			HostEnabled: true,
+			AllowMessages: []string{
+				sdk.MsgTypeURL(&banktypes.MsgSend{}),
```
giansalex, line 74: this message list is outdated, perhaps this section should be removed as it was already included in v11 upgrade

Reecepbcups: @giansalex We add more wasm messages in v12 than were in v11

giansalex: ok, this list should be updated to add v11 msgs (IbcTransfer, MsgUndelegate, etc)

PR #218, opened 2022-05-25
Rename build steps and spike unity prop and gov/sudo testing CI step

the-frey:
This:

- Updates the names of a bunch of CI so it's clearer what it does
- Adds a unity contract CI runner to ensure there are no regressions
- Adds a unity contract governance action to ensure there are no regressions in Juno gov routing to contracts
- Adds a separate docker build _but not push_ step for PRs only (to sense check builds succeed)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,4 +1,4 @@
-name: ci
+name: Build and Push Docker Image
 
 on:
   workflow_dispatch:
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -1,0 +1,26 @@
+name: Build Docker Image on PR
+
+on:
+  pull_request:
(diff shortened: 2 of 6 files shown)


joeabbey (approved):
Nice, preserving unity ... good

faddat:
reading the code now, but I really love the way you're approaching this

.github/workflows/docker.yml
```diff
@@ -1,4 +1,4 @@
-name: ci
+name: Build and Push Docker Image
```
faddat, line 1: can't tell you how many times I've meant to fix naming in ci.  Thanks!

JakeHartnell (approved):

j0nl1:
Dockerfile don't use arch arg anymore and it would be good include arm64 support in the buildx

PR #222, opened 2022-05-30
Integrate ICA Host Functionality

the-frey:
This PR supercedes #215 and seeks to only integrate ICA Host functionality, rather than ICA Controller as well.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -99,7 +99,11 @@ import (
 	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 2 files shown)


the-frey:
@joeabbey @faddat here's a recut of #215 that only integrates ICA Host for now so we can start to test this on uni and in some automated tests.

joeabbey (approved):
Compared side-by-side with #215.

Thanks for paring this down to just Host

faddat (approved):
Like @joeabbey -- I also like how this is just the host, there's a set of odd issues with controller auth.

asalzmann:
Has this been released?

PR #225, opened 2022-06-13
Add store types

the-frey:
As per the IBC go docs https://github.com/cosmos/ibc-go/blob/main/docs/migrations/v2-to-v3.md

I think I missed these in the enabling IBC Host PR, which means we _may_ need this to stop uni falling over.

If I'm wrong, let me know.


the-frey:
@giansalex @joeabbey @faddat @dimiandre I think I missed this when doing the ICA Host enabling so the uni upgrade might fall over tomorrow. Would appreciate a quick scan and review in case I need to get it out to vals ASAP

dimiandre (approved):

PR #207, opened 2022-05-18
✅  remove ignite/starport

faddat:
The ignite cli has a terrible track record of keeping dependencies like `cosmoscmd` up to date.

For security reasons, we should remove any depencency that lives in its repository.

For maintainability reasons, same thing.

New plan is to migrate over the cli tests too.  I think it's possible that they'll help me to identify the issue. 

* removes ignite cli
* bumps to v10
* needed to continue work on:
  * #230 
  * #232  
  * #235 
  * #234

PR #226, opened 2022-06-13
Add dump docker (on failure) to CI

joeabbey:
Looks like the diagnosis of #207 could have gone much faster if we had the logs... trying 

https://github.com/jwalton/gh-docker-logs

(Re-introduced the failure to test)

@faddat I bet this would have saved quite a bit of time:

image


the-frey:
This looks pretty useful for CI purposes

faddat (approved):
I think this is great, and yes, it would have saved tons of time.

the-frey (approved):

PR #229, opened 2022-07-07
Bump SDK version and Juno v7 -> v8

the-frey:
We got stung by this bug during an upgrade:

https://github.com/cosmos/cosmos-sdk/issues/11707

The fix appears to land in `0.45.6` of the SDK - https://github.com/cosmos/cosmos-sdk/pull/12264

So we need to upgrade to `0.45.6`, and increment Juno version from v7 to v8, as this is a consensus breaking change.


faddat:
I will let CI finish then smash the merge button since we've got 3 approving reviews.

PR #230, opened 2022-07-11
DNM: 46

faddat:
This is ready for review by other Juno developers, validators, and the community, but is not yet ready to merge. 

* Upgrades to cosmos-sdk v46
* removes starport
* increments juno's go module version
* adds test that runs a 200s testnet
* enables the command

```
junod testnet start --enable-logging
```

which is probably useful for test contract deployments


the-frey:
Quick CI question 1: why remove gosec?

CI looks like it's failing because of a regression or change that means the contracts can't be stored. Will have to dig into it a bit more.

PR #261, opened 2022-08-08
go 1.19

faddat:
Go 1.19 contains some improvements to the crypto libraries that enhance safety.  I think it makes sense for us to be early adopters. 

https://tip.golang.org/doc/go1.19

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.18
+          go-version: 1.19
       - run: go build ./...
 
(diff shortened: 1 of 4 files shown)


the-frey:
gonna be popular with the valset :)
but yes agree

the-frey (approved):

Highlander-maker:
@the-frey cant we all upgrade to 1.19 now and still run nodes?

the-frey:
@Highlander-maker I would assume 1.19 is compatible with `v9.0.0` yes, but I have not tried it myself yet

faddat:
have tried it; you can

JakeHartnell (approved):

PR #262, opened 2022-08-10
Juno v10

faddat:
@the-frey, @JakeHartnell and I have **talked** a great deal about making faster, incremental releases. 

So here's juno but more special:

* tendermint v0.34.21
* ibc-go v4.0.0
* cosmos-sdk v0.45.7
* iavl v0.19.1

All of these put together make for a leaner, meaner juno. 

## Deploy plan

This is a state breaking change because of ibc-go v4.0.0  It should be tested on Uni first.

## Deploy timeline

We should play around and see how quickly (or not) we can get this out.  The changes are in total quite minor, though they should make juno a good deal faster because we'll be using the Osmosis turbocharged iavl.    

## Addendum

This currently uses a notional-labs fork of wasmd.  I think that this work is now good to go for Uni, but we should wait for this PR:

* https://github.com/CosmWasm/wasmd/pull/940

To be merged upstream before we take this live on mainnet. 

There have been requests from the community to make our block times lower.  When this PR is merged, various validators can/should begin lowering their commit timeout, and we can have faster juno, too.  This is comfortable because of the turbocharged osmosis iavl (now just plain old iavl v0.19.1)


sascha1337:
@faddat i saw some cosmos repos already using ibc go v4(alpha), any estimation if that would introduce breaking changes again? Will check the diff of v3v4 later today

faddat:
definitely agree wiht your changes @the-frey

faddat:
https://github.com/CosmWasm/wasmd/pull/937

faddat:
* @JakeHartnell @joeabbey  @the-frey  hey fyi -- this requires https://github.com/CosmWasm/wasmd/pull/937 to be merged upstream.

faddat:
🔥 

* https://github.com/CosmWasm/wasmd/pull/940

many thanks to Mr. David, another upstream swimmer, for 

* https://github.com/CosmWasm/wasmd/pull/919

the-frey:
Changes look good, I'm just getting caught up now I'm back at my desk. I guess we want to test the blocks/inflation test first and then get this up on uni. Maybe that order.

faddat:
yeah, so I think that this PR should be merged and referred to as the base of Juno v10

Agree that we should do the inflation thing, which likely reqires a gov prop.

Are you referring to the suggestion about making it linear instead of steps, or something else?

the-frey (dismissed):
LGTM, I guess as we move towards an actual v10 RC main will become a working branch until the upstream wasmd changes go in.

faddat:
That's the idea -- also since I figure that Juno is best off as a reference implementation of wasmd, we've also made a wasmd branch based from osmosis-- afaik unfinished.

faddat:
Latest bump brought us current with this wasmd branch:

* https://github.com/CosmWasm/wasmd/pull/940

faddat:
Modified superlinter.yml so that we don't check over go code twice.

faddat:
FYI: this could easily be superceded by:

https://github.com/CosmosContracts/juno/pull/270

They're now both in the same state in that for both PR's we are relying on a Notional fork of wasmd

dimiandre:
> FYI: this could easily be superceded by:
> 
> #270
> 
> They're now both in the same state in that for both PR's we are relying on a Notional fork of wasmd

I'd stick with SDK 45 for now, what do you think @JakeHartnell @the-frey ?

dimiandre (dismissed):
LGTM

faddat:
given events of the past few hours, fully agree @dimiandre

the-frey:
Yep stick with 45

the-frey (approved):

dimiandre (approved):

the-frey:
Leaving a note for self here: version 10 is not ready for production until we can revert to mainline wasmd

PR #272, opened 2022-08-22
Update go.mod

faddat:
Bump back down to go 1.18

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,8 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 # hadolint ignore=DL3018
 RUN set -eux; apk add --no-cache ca-certificates build-base;
 
-RUN apk add git=~2.37.1-r1
+# hadolint ignore=DL3018
+RUN apk add git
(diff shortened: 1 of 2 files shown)


the-frey (approved):
Downgraded the dockerfile as well and resolved the errors with superlinter.

faddat:
Personally I would not have downgraded the docker file but no big deal.

I think that we actually want to build using 19, because that's what most developers are going to be using and stuff like that but we can just bump down the required go version.

I should also probably make us our own build container there's really no reason to use anyone else's and well I'll see what I can do with that.

the-frey:
Well, the docker container should just use the same version as the go mod (if there's a strong disagreement, then it's easy to change that). Kind of disagree about our own build container, base image plus some steps is... kind of fine tbh?

faddat:
Oh no it's not as strong disagreement at all I think we're all good, I was maybe just musing

the-frey:
Cool, let's 🚢 this one then?

faddat:
oh definitely!

dimiandre:
lgtm

dimiandre (approved):

PR #273, opened 2022-08-24
Update mint module

dimiandre:
* Halving time now is based on the total supply instead of fixed blocks per year param.
* Better round block provisions during phase changes

-- 
Need to update tests


nullmames:
I have read through and it appears that the intent from Core discussions has been captured. I have no comment on code.

the-frey:
Had a quick scan. Approach looks good (will read more). Looks like the linter is right, there's some error handling needed?

faddat:
@dimiandre do you mind if I work on this branch directly?

dimiandre:
> @dimiandre do you mind if I work on this branch directly?

I think it's fine! We can then merge and base the other things we have to push on this

faddat:
I think I'd start with the other things first, since they're larger changes, but either way--

Here's a small piece of how we (notional) think of blockchain development-- this PR here is a feature, a deviation from the "base" while 

https://github.com/CosmosContracts/juno/pull/262

is "base".  

We usually go base -> features but I'm good with either direction in this case :).

faddat:
regarding the codeql findings-- basically, the only other way to handle these, is to use a blank import, but the effect is the same: it is a dangerous area of the code.

dimiandre:
> I think I'd start with the other things first, since they're larger changes, but either way--
> 
> Here's a small piece of how we (notional) think of blockchain development-- this PR here is a feature, a deviation from the "base" while
> 
> #262
> 
> is "base".
> 
> We usually go base -> features but I'm good with either direction in this case :).

I prefer your approach more as well, I can rebase this branch on that one. 

Let me know

faddat:
Ok, if you'd like to go that way, basically first step is that we'd get #262 merged, and then we should get an "update branch" option here and shouldn't require any additional lifting beyond that :)

dimiandre:
Testing again on V10, and will start working on upgrade handler.

dimiandre:
All test in my local environment successful. 

Need to write more unit test

faddat:
We've got a team member who does amazing unit tests, I'll see if he has any capacity for this.

x/mint/types/minter_test.go
```diff
@@ -97,15 +94,18 @@ func TestBlockProvision(t *testing.T) {
...
+		totalSupply      sdk.Int
 	}{
-		{secondsPerYear / 5, 1},
-		{secondsPerYear/5 + 1, 1},
-		{(secondsPerYear / 5) * 2, 2},
-		{(secondsPerYear / 5) / 2, 0},
+		{secondsPerYear / 5, 1, sdk.NewInt(1)},
+		{secondsPerYear/5 + 1, 1, sdk.NewInt(1)},
+		{(secondsPerYear / 5) * 2, 2, sdk.NewInt(1)},
+		{(secondsPerYear / 5) / 2, 0, sdk.NewInt(1)},
```
the-frey, line 102: In this test `totalSupply` has no affect on output. I think I would expect adding a new test case like this to fail:

`{(secondsPerYear / 5) * 7, 7, sdk.NewInt(2)},` 

But perhaps I've misunderstood what's going on here :)

x/mint/types/minter_test.go
```diff
@@ -63,25 +63,22 @@ func TestNextPhase(t *testing.T) {
...
-		{101, 1, 1, blocksPerYear, 2},
-		// increase blocks_per_year
-		{102, 2, 101, blocksPerYear * 2, 2},
-		{201, 2, 101, blocksPerYear * 2, 2},
-		{299, 2, 101, blocksPerYear * 2, 2},
-		{300, 2, 101, blocksPerYear * 2, 2},
-		{301, 2, 101, blocksPerYear * 2, 3},
+		{1, 0, 0, blocksPerYear, 1, sdk.NewInt(10000), sdk.NewInt(14000)},
+		{50, 1, 1, blocksPerYear, 1, sdk.NewInt(12000), sdk.NewInt(14000)},
+		{99, 1, 1, blocksPerYear, 1, sdk.NewInt(13960), sdk.NewInt(11400000)},
```
the-frey, line 71: When we say `targetSupply` we mean target supply in this phase only, right?

x/mint/types/minter.go
```diff
@@ -92,7 +93,14 @@ func (m Minter) NextAnnualProvisions(_ Params, totalSupply sdk.Int) sdk.Dec {
 
 // BlockProvision returns the provisions for a block based on the annual
 // provisions rate.
-func (m Minter) BlockProvision(params Params) sdk.Coin {
+func (m Minter) BlockProvision(params Params, totalSupply sdk.Int) sdk.Coin {
```
the-frey, line 96: What is `totalSupply` here?

the-frey:
> Need to write upgrade handler to calculate "TargetSupply"

This bit is done now, right?

x/mint/types/minter_test.go
the-frey: Oh I see, it's because `minter.TargetSupply = tc.totalSupply.Add(minter.AnnualProvisions.TruncateInt())`.

x/mint/types/minter_test.go
dimiandre: yes

x/mint/types/minter_test.go
dimiandre: Yes this test is not very useful with the new method of handling halvings

x/mint/types/minter.go
dimiandre: So, block rewards must be integer. 

With some chain configuration it can happen (and actually is very common) that annual_provision is not a multiple of blocks per year, resulting in a single block reward that would be N ujuno and something. Because 1 ujuno is the minimum amount we can always have, we had to round the result of that division (line 105). 

This leads to phases that can have 1 block more than what it should, by knowing the total supply we can calculate the difference and put it back on track. 

If we don't do that, on mainnet will not be a hug problem, and the maximum supply after 12 years would be max 48 JUNO (4 Juno * phase count) off the target. On testnets or chains where blocks per year is much smaller, therefore single block rewards bigger, the difference is way more noticeable

I can explain better with some example if it's not clear

dimiandre:
> > Need to write upgrade handler to calculate "TargetSupply"
> 
> This bit is done now, right?

yep!

x/mint/types/minter.go
the-frey: Right, that makes sense. Also means that although that parameter is there, it's in most cases going to have _no effect_ on the outcome of the function, right?

x/mint/types/minter_test.go
the-frey: Okay I will add a test case that clarifies that.

x/mint/types/minter_test.go
the-frey: Yep, so I can write a different test to show an unhappy path, but I guess it does show that actually other than in the provision test where we check whether or not to advance a phase, there isn't actually all that much to check. Because it's basically stateless but just driven off current block

x/mint/types/minter.go
dimiandre: that's right! most of the times that check against target supply of the phase is false, happens just once per phase

the-frey: I added a very arbitrary test to show what the effect of that would be in the event it was hit 🤷

giansalex:
what will happen if coins are burned? e.g. whale funds.

the-frey:
@giansalex I guess it will mint more until it hits totalSupply and changes phase? Or is that not right @dimiandre because totalSupply isn't total all-time supply, but instead total supply _in this phase_, in which case it will just continue as it would have done, and the total all-time supply of Juno will decrease.

nullmames:
I guess it depends on if the burn actually removes tokens from bank.

Is our version of a burn sending to a non existing account or are they actually removed from the bank module?

the-frey:
CosmWasm burn, so tokens disappear, removed from circulating supply AFAICR. I will look up the impl later

giansalex:
> I guess it will mint more until it hits totalSupply and changes phase?

yes, but technically more tokens will be minted, burned funds will return to token supply in the future via x/mint module.
loses the point of burning tokens.

the-frey:
Right, so actually you're saying we should calculate the checkpoints ahead of time and stick to them rather than having it function statelessly? Otherwise a burn never actually affects supply in the way intended...

faddat (approved):
Approved, as I think we should deploy a testnet with this.  I also think that we may wish to consider epochs. 

Should we have a call about this today/monday?

the-frey:
I'm now leaning into the opinion that the fact it will make itself back to the correct amount, releasing the difference to stakers is actually... kind of fine? If that changes, we can add additional logic or an upgrade in future to alter.

the-frey (approved):

faddat:
Kind of fine, is imo exactly what we are shooting for here.  I basically won't trust this code until it's been on a testnet, so, lgtm!

<3

(NB: I don't think there's anything wrong with it.  I just want to observe it in its natural habitat some more)

PR #278, opened 2022-09-06
DNM: Experiment with changing compile costs

the-frey:
This is for local testing. Do not merge.


the-frey:
The linter is complaining about a piece of sed wrangling. Can we turn it off for bash scripts? We really don't care.

the-frey:
cc @faddat can we turn off the linter for bash scripts? It's not actually correct in this case for the complex replacement syntax we have. If you change to its suggestion, the substitutions no longer work.

the-frey:
@dimiandre / @faddat this is R4R I think now

dimiandre (approved):
Looks good.

faddat (approved):

faddat (approved):

dimiandre (approved):

PR #279, opened 2022-09-09
Integrate wasmd 0.29.0-rc0

the-frey:
This means going down from ibc-go 4 to ibc-go 3.2. Apart from that, pretty straightforward.


dimiandre (approved):
On mainnet we have v3.0.0 so It will be a minor IBC upgrade, I think it's fine.

faddat:
Needed for now.  Fully agree, though hesitantly.

faddat (dismissed):

faddat:
eep.  We need to change settings so that it doesn't do that when we refresh the branch.  

Approving again.

faddat (approved):

dimiandre (approved):

alpe:
Some nodes:
* you likely want to add `cosmwasm_1_1` to the supportedFeatures in app.go
* A migration is required for IBC v3.0 -> v3.2 
* Please update the docker file to contain the wasmvm 1.1 files

the-frey:
@faddat did we miss the migration here? 
@alpe yep, realized we missed the dockerfile - that's in #282 - but will also add the `cosmwasm_1_1` on that branch... that omission is on me.

the-frey:
Naively I feel like we already did the required migrations as per ICS27 etc from here when we did v7 https://github.com/cosmos/ibc-go/blob/v3.2.1/docs/migrations/v2-to-v3.md

Though I could be wrong...

PR #284, opened 2022-09-20
fix: Ignoring error "validator is not a valid name or address…

deveusss:
This error is fixing error "validator is not a valid name or address: decoding bech32 failed: invalid separator index -1"
When error is thrown, script is unable to call "junod gentx" and "add-genesys-account" functions.
The brings validator into invalid state, and junod returns another error "rpc error: code = NotFound desc = account" while signing transactions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,11 +39,12 @@ RUN chmod +x /opt/*.sh
 
 WORKDIR /opt
 
-# rest server
+## rest server
 EXPOSE 1317
 # tendermint p2p
 EXPOSE 26656
 # tendermint rpc
 EXPOSE 26657
+EXPOSE 9090
 
 CMD ["/usr/bin/junod", "version"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
       - 1317:1317 # rest
       - 26656:26656 # p2p
       - 26657:26657 # rpc
+      - 9090:9090 #grpc
     environment:
       - GAS_LIMIT=${GAS_LIMIT:-100000000}
       - STAKE_TOKEN=${STAKE_TOKEN:-ustake}
--- a/docker/run_junod.sh
+++ b/docker/run_junod.sh
@@ -6,4 +6,4 @@ if test -n "$1"; then
 fi
 
 mkdir -p /root/log
-junod start --rpc.laddr tcp://0.0.0.0:26657 --trace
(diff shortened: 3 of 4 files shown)


the-frey:
This is not an error. It's simply creating a validator if there isn't one. The reason you can't submit a gentx is because the chain is already started. 

If you want to submit a gentx ahead of time you should use a custom genesis file and pass it in. Ditto for the adding of a genesis balance. 

Though, it should be noted for the latter use-case that you can just pass in a different address to the default user and it will be provisioned.

i.e. the address at the end of commands such as those in 

https://docs.junonetwork.io/smart-contracts-and-junod-development/junod-local-dev-setup#run-juno

deveusss:
> This is not an error. It's simply creating a validator if there isn't one. The reason you can't submit a gentx is because the chain is already started.
> 
> If you want to submit a gentx ahead of time you should use a custom genesis file and pass it in. Ditto for the adding of a genesis balance.
> 
> Though, it should be noted for the latter use-case that you can just pass in a different address to the default user and it will be provisioned.
> 
> i.e. the address at the end of commands such as those in
> 
> https://docs.junonetwork.io/smart-contracts-and-junod-development/junod-local-dev-setup#run-juno
--------------------------------------------
this command from docs is also not working for macos.
>This is not an error. It's simply creating a validator if there isn't one.

That the exact problem, it fails when checking for existence)

the-frey:
I've just run this on my machine (mac) and can confirm it works.

```sh
docker run -it \
  --name juno_node_1 \
  -p 1317:1317 \
  -p 26656:26656 \
  -p 26657:26657 \
  -e STAKE_TOKEN=ujunox \
  -e UNSAFE_CORS=true \
  ghcr.io/cosmoscontracts/juno:v9.0.0 \
  ./setup_and_run.sh juno16g2rahf5846rxzp3fwlswy08fz8ccuwk03k57y
```

the-frey:
Also confirmed working on MBP 2018 with Intel

the-frey:
When you run 

```sh
docker run -it \
  --name juno_node_1 \
  -p 1317:1317 \
  -p 26656:26656 \
  -p 26657:26657 \
  -e STAKE_TOKEN=ujunox \
  -e UNSAFE_CORS=true \
  ghcr.io/cosmoscontracts/juno:v9.0.0 \
  ./setup_and_run.sh juno16g2rahf5846rxzp3fwlswy08fz8ccuwk03k57y
  ```
  
What is the error you see?

PR #289, opened 2022-09-23
ledger patch

faddat:
This replaces the 99designs keyring with the cosmos keyring fork of the 
same lib


the-frey:
You need to rebase this on top of the rc-1 branch if you want to use rc-1, because a bunch of stuff has changed :)

faddat:
no need to fix that which isn't broken.

faddat:
no need to fix that which isn't broken.

PR #292, opened 2022-09-26
DNM: Experiment around with CI

the-frey:


the-frey:
This *looks* like 5_000_000 might be too low as block gas. Needs further investigation.

PR #298, opened 2022-10-02
Update readme.md

dimiandre:


nullmames:
LGTM

the-frey:
This is good, but I've been thinking for ages we should just bin loads of the stuff that's just not relevant. Only devs and validators read this page (post genesis) - have suggested a slightly edited and much cut-down version in #299

faddat:
I think I prefer @the-frey more reductive approach.  Also want to add a state sync step, too.

faddat:
I think I prefer @the-frey more reductive approach.  Also want to add a state sync step, too.

PR #301, opened 2022-10-03
DNM : Liquidity staking module

vuong177:
This PR is a example for using the liquidity staking module in Juno.
In this PR, I use sdk46 and ibc v5 because iqlusioninc team use sdk46 in their module.


faddat:
@vuong177 nice work here!

This PR shows pretty well, the issue that I've hit each time trying to 46-ify Juno, specifically I think that we need to update CI to reflect the cli changes in 46.

faddat:
Hey, so I guess the question here is weather we refresh, or wait till sdk 47.

vuong177:
@faddat it's an example. we will not use this !

PR #303, opened 2022-10-03
bump to v11

vuong177:
Change import paths to v11


faddat:
@vuong177 to keep this PR small, I changed the description for it to match what it currently does. 

Nice work!

faddat:
hey uh, im sorry, I accidentally edited your reply.

faddat (approved):

the-frey:
I think main can now be merged into this. Superlinter is disabled on main, and then there's an upgrades folder with a draft handler for v11 (wasmd has a couple changes that need adding to ICA) - so that handler can just be called.

faddat:
@the-frey I reckon we can close this now?  I just merged these changes in the form of #313

PR #312, opened 2022-10-08
Reset application.db and  wasm

chillyvee:
Is there a better way to reset application.db and wasm?  For example by extending into the "tendermint" reset-state command?

### DIFF
--- a/cmd/junod/tendermint.go
+++ b/cmd/junod/tendermint.go
@@ -50,12 +50,12 @@ func resetWasm(dbDir string) error {
 		if err := os.RemoveAll(wasmDir); err == nil {
 			fmt.Println("Removed wasm", "dir", wasmDir)
 		} else {
-			fmt.Println("error removing wasm", "dir", wasmDir, "err", err)
+			return fmt.Errorf("error removing wasm  dir: %s; err: %w", wasmDir, err)
 		}
 	}
 
 	if err := tmos.EnsureDir(wasmDir, 0700); err != nil {
-		fmt.Println("unable to recreate wasm", "err", err)
+		return fmt.Errorf("unable to recreate wasm %w", err)
 	}
 	return nil
 }
@@ -68,12 +68,12 @@ func resetApp(dbDir string) error {
 		if err := os.RemoveAll(appDir); err == nil {
 			fmt.Println("Removed application.db", "dir", appDir)
 		} else {
-			fmt.Println("error removing application.db", "dir", appDir, "err", err)
+			return fmt.Errorf("error removing application.db  dir: %s; err: %w", appDir, err)
 		}
 	}
 
 	if err := tmos.EnsureDir(appDir, 0700); err != nil {
-		fmt.Println("unable to recreate application.db", "err", err)
+		return fmt.Errorf("unable to recreate application.db %w", err)
 	}
 	return nil
 }
--- a/cmd/junod/tendermint.go
+++ b/cmd/junod/tendermint.go
@@ -14,8 +14,8 @@ import (
 
 // ResetWasmCmd removes the database of the specified Tendermint core instance.
 var ResetWasmCmd = &cobra.Command{
-	Use:   "reset-app-wasm",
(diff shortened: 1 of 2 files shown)


the-frey:
This looks potentially useful.

Would also say that it's worth maybe removing cache as well in a separate command 🤔 

There have been a bunch of changes on main though so might be easier to start fresh and just add these changes (or rebase)

chillyvee:
Can take another look soon since testnet completed the upgrade.

chillyvee:
> Would also say that it's worth maybe removing cache as well in a separate command 🤔

Split the `application.db` and `wasm` reset commands

Rebased on main

cmd/junod/tendermint.go
```diff
@@ -0,0 +1,79 @@
...
+
+// resetWasm removes wasm files
+func resetWasm(dbDir string) error {
+	wasmDir := filepath.Join(dbDir, "wasm")
+
+	if tmos.FileExists(wasmDir) {
+		if err := os.RemoveAll(wasmDir); err == nil {
+			fmt.Println("Removed wasm", "dir", wasmDir)
+		} else {
+			fmt.Println("error removing wasm", "dir", wasmDir, "err", err)
```
vuong177, line 53: return fmt.Errorf("error removing wasm %s dir: err %w", wasmDir, err)

cmd/junod/tendermint.go
```diff
@@ -0,0 +1,79 @@
...
+
+// resetApp removes application.db files
+func resetApp(dbDir string) error {
+	appDir := filepath.Join(dbDir, "application.db")
+
+	if tmos.FileExists(appDir) {
+		if err := os.RemoveAll(appDir); err == nil {
+			fmt.Println("Removed application.db", "dir", appDir)
+		} else {
+			fmt.Println("error removing application.db", "dir", appDir, "err", err)
```
vuong177, line 71:
```suggestion
			return fmt.Errorf("error removing application.db  dir: %s; err: %w", wasmDir, err)

File: cmd/junod/tendermint.go (hunk 1 above)
Comment (vuong177) [inline cmd/junod/tendermint.go:53]:
```suggestion
			return fmt.Errorf("error removing wasm  dir: %s; err: %w", wasmDir, err)
```

cmd/junod/tendermint.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+
+	if err := tmos.EnsureDir(appDir, 0700); err != nil {
+		fmt.Println("unable to recreate application.db", "err", err)
```
vuong177, line 76: same nits
```suggestion
		return fmt.Errorf("unable to recreate application.db %w", err)
```

cmd/junod/tendermint.go
```diff
@@ -0,0 +1,79 @@
...
+	if tmos.FileExists(wasmDir) {
+		if err := os.RemoveAll(wasmDir); err == nil {
+			fmt.Println("Removed wasm", "dir", wasmDir)
+		} else {
+			fmt.Println("error removing wasm", "dir", wasmDir, "err", err)
+		}
+	}
+
+	if err := tmos.EnsureDir(wasmDir, 0700); err != nil {
+		fmt.Println("unable to recreate wasm", "err", err)
```
vuong177, line 58: same nits
```suggestion
		return fmt.Errorf("unable to recreate wasm %w", err)
```

vuong177 (changes requested):

vuong177:
@chillyvee, thanks for making this. Only some minors left, I will approve when it is addressed.

chillyvee:
@vuong177 Thank you for the feedback.  We'll get this amended as soon as we have a chance.

chillyvee:
@vuong177 Your suggestions have been added to the branch

Reecepbcups (approved):
Approved but my only thought is should we have this in the root commands?
or make its own reset section? `junod reset app`, `junod reset wasm` @chillyvee 

In the future we may want to add resetting config files like app.toml as well to default

PR #320, opened 2022-10-15
Set IAVL config

giansalex:
the configuration can be changed in `.juno/config/app.toml`

```toml
# IavlCacheSize set the size of the iavl tree cache.
# Default cache size is 50mb.
iavl-cache-size = 781250

# IAVLDisableFastNode enables or disables the fast node feature of IAVL.
# Default is true.
iavl-disable-fastnode = true

```

perhaps it should be added to release notes for next version


faddat (approved):
full approve.  Thank you!

the-frey (approved):
Good shout on the notes as well - I'll add a reminder to put that in.

PR #328, opened 2022-10-24
x/globalfee

### LINKED ISSUE #327: Global fee module [v46/v47 sdk]
**note: postponing until juno moved to v46/47 so we can import from gaia directly**

A way to set a minimum global fee via governance param for different tokens, this is NOT going to be in SDK47

Ex: force all nodes to use 0.025ujuno as a fee. 
Can use as many token denoms here as you want (including ibc). 

Default genesis is empty array, param change to update
If the fee token is less than the globalfee, the Tx fails for not enough fee

Reecepbcups:
Closes #327  for v12 (nice to have)

`junod q globalfee minimum-gas-prices` & require minimum fees for Txs


the-frey:
Question: given this is landing in the next version of the SDK is it worth us including this code in Juno?

Makefile/proto changes are def a good developer convenience though - totally onboard with that and think we should stick that in another PR and ship it, like now 👍

the-frey:
Wait - before you do @faddat I'm not imagining things right? That is a 47 feature?

Reecepbcups:
https://github.com/cosmos/cosmos-sdk/issues/12269#issuecomment-1272646147

Seems they could include it, but does not seem it will actually be (as of 18 days ago)

vuong177 (changes requested):
@Reecepbcups  We need to create an upgrade handler for this.

Reecepbcups:
I feel we should wait for v47 then we can just import the Gaia fork directly with native fee bypass as well (for IBC)

NOTE: move proto work -> It’s own PR

Reecepbcups:
Jack added we should write this in a way a Committee (ex: security DAO / 3/5sig) can change prices for X amount of time without the need for a proposal. In case of an attack. Then once used to raise fees, dao has to make a proposal to 'top up' their ability to do so.

If this is the case we should likely just fork the gaia as originally intended & add these features

PR #351, opened 2022-11-02
fix(Dockerfile): latest alpine, use sh, combine run

pratikbin:
- Use latest alpine based golang image
- Use sh just to standardized shells
- combine `RUN` instructions
- Use `entrypoint` to handle os signals

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,15 @@
 # docker build . -t cosmoscontracts/juno:latest
 # docker run --rm -it cosmoscontracts/juno:latest /bin/sh
-FROM golang:1.18-alpine3.15 AS go-builder
+FROM golang:1.18-alpine AS go-builder
 
 # this comes from standard alpine nightly file
 #  https://github.com/rust-lang/docker-rust-nightly/blob/master/alpine3.12/Dockerfile
 # with some changes to support our toolchain, etc
-SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
+SHELL ["/bin/sh", "-ecuxo", "pipefail"]
 # we probably want to default to latest and error
 # since this is predominantly for dev use
 # hadolint ignore=DL3018
-RUN set -eux; apk add --no-cache ca-certificates build-base;
-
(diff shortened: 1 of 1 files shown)


faddat (approved):

faddat:
why change to sh instead of ash?

either way looks good to me

pratikbin:
> why change to sh instead of ash?
> 
> either way looks good to me

It's just to standardize the shell

faddat (approved):

the-frey:
Triggered workflows. Let's see what CI says 🎲

Dockerfile
```diff
@@ -53,4 +50,5 @@ EXPOSE 26656
 # tendermint rpc
 EXPOSE 26657
 
-CMD ["/usr/bin/junod", "version"]
\ No newline at end of file
+ENTRYPOINT ["/usr/bin/junod"]
```
the-frey, line 53: This will be a breaking change for consumers of this that assume the current run semantics, right? as in, they will have to change the command to omit the first part?

the-frey: Sense checking, as AFAICR all checks that use docker in CI rely on `exec` semantics, which will ignore any change to entrypoint

pratikbin: > This will be a breaking change for consumers of this that assume the current run semantics, right? as in, they will have to change the command to omit the first part?

true. I realize that after checking CI

the-frey: Yeah, I think on that basis we have to maintain existing API just since folks will be relying on it in build scripts and all sorts of dev scenarios.

pratikbin: > 

Let me revert it, since it can break existing folks running this

faddat: thanks

faddat (approved):

the-frey (approved):

the-frey:
Hm, although, why's it only showing 9 CI checks when other PRs are 11?

pratikbin:
code-ql only runs on go file changes and tidy and test on push.

PR #360, opened 2022-11-09
Update genica command

the-frey:
The messages in genica command are now out-of-date, since v11 has shipped.

This simply... does that.

To confirm - check the messages added in upgrade handler match the ones added here.


faddat:
thank you!

Issue #245, opened 2022-08-05
Cleaner juno/app


Anmol1696:
## Overview

Create `app/upgrades` dir to have all the upgrades in one place with versions and abstract out all keepers to `app/keepers` to make the the `app/app.go` cleaner and more maintainable. 

Problem Definition
app/app.go is too big file.

Proposal
split app/app.go into app.go, upgrade/, keepers/ , module.go

ref:
https://github.com/cosmos/gaia/issues/1557
https://github.com/cosmos/gaia/pull/1580 
[osmosis/app](https://github.com/osmosis-labs/osmosis/tree/main/app)
[provenance/app](https://github.com/provenance-io/provenance/blob/main/app/upgrades.go)

faddat:
Hi there! I really liked your PR for the hub, this one:

* https://github.com/cosmos/gaia/pull/1580

I have been in seoul and unable to give it a proper review.  

Are the osmo and provenance styles different?

Point blank: I agree with you.

sascha1337:
@faddat - refactor useful? might be some good practice to get warm

Anmol1696:
@faddat provenance style is a little different in the sense they define there upgrades as a map with values as the upgrade function like [this](https://github.com/provenance-io/provenance/blob/main/app/upgrades.go#L29), then they have functions to deal with this map and create handlers out of this. 

In osmo this is more modular and we create a package itself per upgrade version, instead of a key value in map.

I prefer the osmo way since it is cleaner and will be able to do alot more, includig handling forks better as well

faddat:
I'm very happy ot use the osmo style :D

PR #366, opened 2022-11-12
app.go refactor: Move keepers init, modules init and modularise upgrade handlers

udit-gulati:
Closes https://github.com/CosmosContracts/juno/issues/245

I'm a bit unsure about upgrade handlers cause in existing `app.go` there is upgrade name check for  `"multiverse"` but there are two upgrades (v10 and v11), so not sure which one is which.


udit-gulati:
@faddat yes, the upgrade handlers are osmosis style and remaining changes are the same style in osmosis and gaia.

faddat (approved):

Reecepbcups (approved):

PR #368, opened 2022-11-14
TokenFactory (v12)

### LINKED ISSUE #334: Token Factory Convert Tools
Dec-7-2022 meeting notes: https://hackmd.io/@reecepbcups/cw20-to-tokenfactory

In many ways a lot better than current CW20 implementation, especially with bank query and Developer UX
_Needed for UltraSoundJuno project as well_
<br/>

Apart of this, we should write scripts / documentation on converting CW20s to the native factory.
- [Airdrop Style](https://github.com/Reecepbcups/cw20-to-native)
- [CW20 Burn Contract (send from balance & mint options)](https://github.com/Reecepbcups/cw20-tokenfactory-burn)
- CW20 impl the tokenfactory interface (native bank query and mint)
- others?
- https://github.com/CosmosContracts/tokenfactory-contracts
<br/>

- Question is best approach for JunoSwap as well for already public CW20 tokens if they do move to native

- Taking into account dao governance CW20 tokens

Reecepbcups:
Close #334

- adds back in tokenfactory with the new main format
- adds v12 upgrade handler for tokenfactory with the new format


Reecepbcups:
@faddat Do we want to merge this in now since it also has the v12 upgrade handler? If so only thing I need to do is test this with a wasm contract to ensure it is good

faddat:
Yes

faddat (dismissed):

app/upgrades/v12/constants.go
```diff
@@ -0,0 +1,18 @@
+package v11
```
udit-gulati, line 1:
```suggestion
package v12
```

app/upgrades/v12/constants.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	tokenfactorytypes "github.com/CosmWasm/token-factory/x/tokenfactory/types"
+	"github.com/CosmosContracts/juno/v12/app/upgrades"
+	store "github.com/cosmos/cosmos-sdk/store/types"
+)
+
+// UpgradeName defines the on-chain upgrade name for the Juno v11 upgrade.
```
udit-gulati, line 9:
```suggestion
// UpgradeName defines the on-chain upgrade name for the Juno v12 upgrade.
```

app/upgrades/v12/upgrades.go
```diff
@@ -0,0 +1,19 @@
+package v11
```
udit-gulati, line 1:
```suggestion
package v12
```

Reecepbcups: Thank you, thought I had got this but seems I overlooked it

app/keepers/keepers.go
```diff
@@ -62,7 +66,7 @@ type AppKeepers struct {
 
 	// keepers
 	AccountKeeper    authkeeper.AccountKeeper
-	BankKeeper       bankkeeper.Keeper
+	BankKeeper       bankkeeper.BaseKeeper
```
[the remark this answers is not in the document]
Reecepbcups: Its a requirement for the TokenFactory wasm bindings -> https://github.com/CosmWasm/token-factory/blob/1345f322c8877322d974c316d46c0b0aa6f24e80/x/tokenfactory/bindings/wasm.go#L12-L15

app/upgrades/v12/upgrades.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// CreateV12UpgradeHandler makes an upgrade handler for v12 of Juno
+func CreateV12UpgradeHandler(
+	mm *module.Manager,
+	cfg module.Configurator,
+	keepers *keepers.AppKeepers,
+) upgradetypes.UpgradeHandler {
+	return func(ctx sdk.Context, _ upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+		return mm.RunMigrations(ctx, cfg, vm)
```
vuong177, line 17: you should set token factory param in here.

Reecepbcups:
@the-frey / @vuong177  Do we wan't a denomination creation fee? 

I set to 1 JUNO here but maybe we make it free to do so or like 0.01 juno just to prevent spam creation of denoms

vuong177:
I think we should set it to 1 JUNO or higher.

Reecepbcups:
@the-frey Currently someone could spam upload CW20s for free. Given the bank module has better queries I do feel there is less overhead overall for a chain.

Most developers will need at minimum 2 tokens most likely, one for messing around first & one for actual production. I feel keeping the barrier to entry really low to start is best course of action.

If we find lots of spam, we can gov prop to increase param as needed

the-frey:
Yep, happy with 1JUNO as a nominal fee

faddat (approved):
The road to uni

faddat (approved):

vuong177 (approved):

Issue #371, opened 2022-11-17
[CI] Update GitHub Action to push docker images


anhductn2001:
## Background

 Updates the GitHub Actions CI to push docker images due to the deprecation of set-output commands in Github Action (source: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

The functionality's still the same.

## Suggested Design

 Update GitHub action logic
 Rename it to push-docker-images.yaml

## Acceptance Criteria

- all existing and new tests should pass

faddat:
great idea

PR #363, opened 2022-11-09
E2E

vuong177:
## What is the purpose of the change

 Add End to end tests for Juno.
 The e2e package defines an integration testing suite used for full end-to-end testing functionality.
 It initializes the chains for testing via Docker.
### Features included in this PR
1. Basic logic

* This is the most basic type of setup where a single chain is created
* It simply spins up the desired number of validators on a chain.
3. Test IBC Token Transfer
* Ibc transfer test
* Using Hermes Relayer
4. Test Token Factory Bindings
* TokenFactory module and its bindings work as expected
5. Upgrade testing
* CLI commands are run to create an upgrade proposal and approve it
* Old version containers are stopped and the upgrade binary is added
* Current branch Juno version is spun up to continue with testing
### Features will do in future

 - State Sync Testing
 - TestTWAP
 - Test Global Fee

PR #383, opened 2022-11-27
e2e - tokenfactory bindings test

### LINKED ISSUE #377: Token factory testing
## Background
- Following the token factory module being added to JUNO, we need more testing for this 

## Suggested Design
-  Have a test contract and test its interaction with the token factory module

Reecepbcups:
Closes #377 
Merges into e2e  branch (not main). Does not require approvals

- Ensures the wasm bindings works with a real contract, tested with e2e testing
- tested locally with `make test-e2e`
- e2e passes now in this branch


vuong177:
hey @Reecepbcups  test-e2e still fail.

Reecepbcups:
@vuong177 -> https://github.com/CosmosContracts/juno/pull/363#issuecomment-1319618904
Since I built off e2e this is why it fails. Will be fixed when this merges -> e2e-test -> main

vuong177:
@Reecepbcups this issue https://github.com/CosmosContracts/juno/issues/371 for automate build docker image when we release  new version. I think e2e test should be passed in this PR.

vuong177:
@Reecepbcups I also need to review tf contract code. Did you create a repo for it.

Reecepbcups:
> @Reecepbcups I also need to review tf contract code. Did you create a repo for it ?

@vuong177 https://github.com/CosmWasm/token-bindings/tree/main/contracts/tokenfactory
osmosis's example contract was compiled and used

Reecepbcups:
I have not touched .github folder, but above `every commit / test` fails for 
`API error (404): pull access denied for juno, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

vuong177:
Need to seperate normal test and e2e test when run CI. Add a go tag could be fix this.

tests/e2e/e2e_test.go
```diff
@@ -1,15 +1,21 @@
-//go:build e2e
```
vuong177, line 1: hey why you delete this ? Test fail because you remove this line

vuong177 (changes requested):

tests/e2e/e2e_test.go
Reecepbcups: Whoops! This was from when I was trying to get it to run locally. Reverting back to how it was before sorry!

tests/e2e/e2e_setup_test.go
```diff
@@ -100,7 +99,10 @@ func (s *IntegrationTestSuite) SetupSuite() {
 
 	if str := os.Getenv(upgradeVersionEnv); len(str) > 0 {
 		upgradeSettings.Version = str
-		s.T().Log(fmt.Sprintf("upgrade version set to %s", upgradeSettings.Version))
+		s.T().Logf("upgrade version set to %s", upgradeSettings.Version)
+	} else {
+		upgradeSettings.Version = "v12"
```
vuong177, line 104: `upgradeSettings.Version` shouldn't be hardcoded

vuong177:
@Reecepbcups hmm I don't see the upgrade test in e2e log.

Reecepbcups:
> @Reecepbcups hmm I don't see the upgrade test in e2e log.

@vuong177 The v12 upgrade, deposit, and vote test? Done here -> https://github.com/CosmosContracts/juno/actions/runs/3558650155/jobs/5977470658#step:9:563
upgrades to height 70 then the token binding test start & all are successful at the end

Since its json, success check is `"\"code\":0,"` now instead of `code: 0` given all logs were that way

vuong177:
LGTM

PR #384, opened 2022-11-29
Minor: Do not need push image when merge main

anhductn2001:


JakeHartnell (approved):
I'm ok with this.

vuong177 (approved):

the-frey:
Not needed because there will be a build that corresponds to the commit when it hits main?

anhductn2001:
In my opinion, image should only be built when we release a version. Changes to the current main can be true or false. For debugging or testing, developers can build their own image to debug with `make docker-build-debug` or `docker-compose up` in local

PR #385, opened 2022-11-30
implement x/feeshare

Reecepbcups:
This is the module to fulfill the % of fees going to dAPP developers. 

**v12 maybe?? **

Features:
- You can specify a separate withdraw address
- Only contracts you are an admin of can you register payouts for (opt in requirement for tax reasons)
- 50% of funds by default, governance can enable / change the % to any amount they like

Needs:
- testing + e2e
- better fee payout logic (ex: split the 50% fees between contracts which opted in to register for payout)
- Remove testing files "test_node.sh, test_fee.sh, cw_template.wasm, cw20_base.wasm"


scripts/protocgen.sh
```diff
@@ -26,7 +26,7 @@ cd ..
 buf generate
 
 # move proto files to the right places
-cp -r ./github.com/CosmosContracts/juno/x/* x/
+sudo cp -r ./github.com/CosmosContracts/juno/x/* x/
```
JakeHartnell, line 29: Feels weird... are we sure this is needed?

x/feeshare/README.md
```diff
@@ -0,0 +1,6 @@
+# fee-share
```
JakeHartnell, line 1: Would be nice to have a little more documentation for a custom module such as this. :heart:

Reecepbcups: Resolved -> https://github.com/CosmosContracts/juno/pull/385/commits/91d60e422c49c3567535df0460a3cc3f9e0a387f

GNaD13 (commented):
I think we still have some minor issue and not have proposal for change DeveloperShares yet. Can you check this?

x/feeshare/ante/ante.go
```diff
@@ -0,0 +1,119 @@
...
+// FeePayLogic takes the total fees and splits them based on the governance params
+// and the number of contracts we are executing on.
+// This returns the amount of fees each contract developer should get.
+// tested in ante_test.go
+func FeePayLogic(fees sdk.Coins, govPercent sdk.Dec, numPairs int) sdk.Coins {
+	var splitFees sdk.Coins
+	for _, c := range fees {
+		rewardAmount := govPercent.MulInt(c.Amount).QuoInt64(int64(numPairs)).RoundInt()
+		if !rewardAmount.IsZero() {
+			splitFees = append(splitFees, sdk.NewCoin(c.Denom, rewardAmount))
```
GNaD13, line 54: Should add check if c.Denom already have in splitFees. If so, should addition rewardAmount with the current amount of c.Denom in splitFees

Reecepbcups:
> I think we still have some minor issue and not have proposal for change DeveloperShares yet. Can you check this?

@GNaD13 It is the the params, so `junod tx gov submit-proposal param-change` would do

x/feeshare/ante/ante.go
Reecepbcups: https://github.com/CosmosContracts/juno/pull/385/commits/69087122eabfdfb534edbe651207cc4078298418

faddat:
I am now feeling lgtmification here.

Makefile
```diff
@@ -17,6 +17,7 @@ LEDGER_ENABLED ?= true
 SDK_PACK := $(shell go list -m github.com/cosmos/cosmos-sdk | sed  's/ /\@/g')
 TM_VERSION := $(shell go list -m github.com/tendermint/tendermint | sed 's:.* ::') # grab everything after the space in "github.com/tendermint/tendermint v0.34.7"
 DOCKER := $(shell which docker)
+DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bufbuild/buf:1.0.0-rc8
```
faddat, line 20: nice

faddat (approved):

app/modules.go
```diff
@@ -118,9 +121,10 @@ func appModules(
 		authzmodule.NewAppModule(appCodec, app.AuthzKeeper, app.AccountKeeper, app.BankKeeper, app.interfaceRegistry),
 		transfer.NewAppModule(app.TransferKeeper),
 		ica.NewAppModule(nil, &app.ICAHostKeeper),
+		tokenfactory.NewAppModule(app.TokenFactoryKeeper, app.AccountKeeper, app.BankKeeper),
+		feeshare.NewAppModule(app.FeeShareKeeper, app.AccountKeeper),
 		// this line is used by starport scaffolding # stargate/app/appModule
```
JakeHartnell, line 126:
```suggestion
```

Probably can kill this very old comment.

x/feeshare/README.md
```diff
@@ -0,0 +1,8 @@
...
+
+This module shares contract execution fees with the smart contract developers as passed in [Juno Prop 51](https://www.mintscan.io/juno/proposals/51)
+
+This module is a heavily modified fork of [evmos/x/revenue](https://github.com/evmos/evmos/tree/main/x/revenue)
```
JakeHartnell, line 5:
```suggestion
This module is a heavily modified fork of [evmos/x/revenue](https://github.com/evmos/evmos/tree/main/x/revenue).
```

x/feeshare/ante/ante_test.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func TestAnteSuite(t *testing.T) {
+	suite.Run(t, new(AnteTestSuite))
+}
+
+func (suite *AnteTestSuite) TestFeeLogic() {
+
+	// We expect all to pass
+	feeCoins := sdk.NewCoins(sdk.NewCoin("ujuno", sdk.NewInt(500)))
```
JakeHartnell, line 23: Are there any tests for other denoms or fees payed in multiple tokens?

Reecepbcups: We only accept juno as payout per the ante (tax reasons alex brought up monday)

x/feeshare/keeper/msg_server.go
```diff
@@ -0,0 +1,226 @@
...
+	deployerAddr, err := sdk.AccAddressFromBech32(deployer)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid deployer address %s", deployer)
+	}
+
+	info := k.wasmKeeper.GetContractInfo(ctx, contract)
+
+	if len(info.Admin) == 0 {
+		return nil, sdkerrors.Wrapf(sdkerrors.ErrUnauthorized, "contract %s has no admin set", contract)
+	}
```
JakeHartnell, line 26: Ok, have one discussion point here which I would be curious to get folks thoughts on.

IMO, this should be contract `admin` _or_ `creator` (only if `admin` is not set). Sometimes contracts don't need an `admin`, and as `admin` on contracts has the potential to be abused through contract migrations I don't think we should be incentivizing contracts with admin exclusively.

However, I do think `admin` should take precedence, so would only allow for `creator` to call something like `MsgRegisterFeeshare` if an `admin` is not set.

Thoughts @Reecepbcups? Others?

JakeHartnell: Comment is probably more relevant for line 58 below: https://github.com/CosmosContracts/juno/pull/385/files#diff-81a0948bc3c50667477e54062864d6fa9d845d0d428ea90ffc0386f6ef7a539fR58

Reecepbcups: @JakeHartnell I originally had the logic for creators to have control if there was no admin set. My is that you can't change the `creator`, but you can change the `admin`. But this actually doesn't apply since the whole goal is to not change the person in control. So that is my bad. 

I think it makes sense to add this rather than force admin contracts for fees, which is bad as you alluded too

Reecepbcups: https://github.com/CosmosContracts/juno/pull/385/commits/70d48bbb50397e1ee2218267a7b3d55d3d4ca7ea

JakeHartnell (approved):
Amazing work! Can't wait to play with this in the next testnet!

the-frey:
LGTM. I think we'll end up with some nits once we move this into test, but we probably won't catch them all until we get it on uni and start kicking the tyres.

faddat (approved):

PR #390, opened 2022-12-03
Use go 1.19

faddat:
This PR responds to validator feedback by mandating the use of go 1.19.  Our next state breaking change (v13) is a good time to do this and we will then be able to ensure that all validators are using the same version of Go when they build the binary.


Reecepbcups (approved):
Agree this is a good change for v12

jackzampolin (approved):

PR #391, opened 2022-12-03
tendermint v0.34.24

faddat:
This PR brings Juno current with the latest release of tendermint.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/spf13/cast v1.5.0
 	github.com/spf13/cobra v1.6.1
 	github.com/stretchr/testify v1.8.1
-	github.com/tendermint/tendermint v0.34.23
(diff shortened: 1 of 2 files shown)


Reecepbcups:
@faddat Do we have to talk with the Skip team before doing such upgrades to tendermint? Since they maintain a custom version. Didn't know if this could conflict & cause extra headache for them

jackzampolin (approved):

faddat:
@Reecepbcups -- since this is for v12, I think this is skip-fine. 

I'll also update in an issue about skip overall, I have ideas, very similar to what Dimi mentioned.

jackzampolin (approved):

Reecepbcups:
Talked with Barry tonight, TM upgrades are fine in the future. 

Was a one off proto issue between versions (state breaking for Skip protocol, but not underlying TM). So we should be good

PR #395, opened 2022-12-04
Remove ibc-fee

vuong177:
This PR removes the dependency on ibc-fee from the wasmd-bump branch.


Reecepbcups:
No longer doing, using IBCFee for v12

PR #401, opened 2022-12-08
bech32 convert command

faddat:
bech32 conversions should be a first class citizen, as it is a very useful tool


cmd/junod/cmd/root.go
```diff
@@ -108,7 +107,7 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
 		AddGenesisWasmMsgCmd(app.DefaultNodeHome),
 		tmcli.NewCompletionCmd(rootCmd, true),
 		// testnetCmd(app.ModuleBasics, banktypes.GenesisBalancesIterator{}),
-		debug.Cmd(),
+		ConvertBech32Cmd(),
```
Reecepbcups, line 110: Why is the `junod debug` command removed with this? This can still be added on top of the debug.Cmd()

faddat: sensible!

Reecepbcups:
Feel its better to not spam/fill up the root cmd with random tools

Reecepbcups (approved):

faddat:
unsure about this but it is nbd so lfg

JakeHartnell (approved):
LGTM!

PR #402, opened 2022-12-08
bench32.go should live in cmd/

anhductn2001:


faddat:
hey, how is this different from #401

anhductn2001:
It is same. I just put the file in the correct directory because it gives an error in Github Action

faddat:
oh it is **into** 401.  I missed that.  Thanks!

PR #404, opened 2022-12-08
Faddat/mev experiment

faddat:
This is for Juno validators using tendermint-mev **to test outside their validator nodes**

* you do not need to be using Skip in order to use this work
* to integrate this on your validator node, you'd simply run this binary
* the p2p logs will look different, because I've changed them in this:
  * https://github.com/notional-labs/mev-tendermint/tree/v0.34.x-mev-gadikian 
* speaking as myself only, I'm interested in seeing this become Juno's canonical binary, regardless of weather validators are using Skip or not.

```bash
bash scripts/statesync-mev.bash
```

## More context

The tendermint core repository is reather difficult to ship improvements to, this PR is going straight to the point:

* only pebbledb is supported for tendermint's store
* cosmos-sdk still supports goleveldb and maybe others but I think that the 5x reduction in RAM usage seen on Osmosis archive nodes when using pebble -- and associated big reductions in complexity is a great reason to take these choices away from infrastructure operators and instead use the clear sane default.
* Notional's validator will soon use this build in order to prove its viability.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/spf13/cast v1.5.0
 	github.com/spf13/cobra v1.6.0
 	github.com/stretchr/testify v1.8.0
-	github.com/tendermint/tendermint v0.34.23
+	github.com/tendermint/tendermint v0.34.24
 	github.com/tendermint/tm-db v0.6.7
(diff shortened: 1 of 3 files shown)


faddat:
This was an experimental branch and I think it can now safely be closed.

PR #411, opened 2022-12-12
x/globalfee

Reecepbcups:
Pulled forward from Gaiav8 - SDK v45 for Juno v12
Merges wasmd 30 branch into here

Commonwealth: https://commonwealth.im/juno/discussion/8386-implement-xglobalfee

Feature:
```
- Sets a minimum fee for the entire network
- A Node can set a higher fee than the minimum, but not lower
- Validators can not accept anything which is not in the globalfee params
```

After upgrade, community can decide on good fees / denoms to accept (usdc, juno, osmo)

`junod tx gov submit-proposal param-change ~/Desktop/Work/paramChange.json --from $KEY --keyring-backend os --chain-id juno-1 --yes --broadcast-mode block`
```json
{
    "title": "MinimumGasPricesParam Change",
    "description": "add juno minimum fee",
    "changes": [
      {
        "subspace": "globalfee",
        "key": "MinimumGasPricesParam",
        "value": [{"denom":"ujuno","amount":"0.002500000000000000"}]
      }
    ],
    "deposit": "1000000ujuno"
}
```

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -31,9 +31,12 @@ import (
 	govclient "github.com/cosmos/cosmos-sdk/x/gov/client"
 	paramsclient "github.com/cosmos/cosmos-sdk/x/params/client"
 	paramstypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	upgradeclient "github.com/cosmos/cosmos-sdk/x/upgrade/client"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 	ibcclientclient "github.com/cosmos/ibc-go/v3/modules/core/02-client/client"
+	ibcclienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
+	ibcchanneltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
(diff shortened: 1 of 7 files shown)


Reecepbcups:
@faddat built off main, and used the same as before. Works just fine... If you could update the commonwealth to this PR now. I closed the other. Ready for review

the-frey (approved):
LGTM 👍 

Weird about that other thing.

faddat (approved):

faddat:
@Reecepbcups - stuff's odd sometimes (re: rebuilding off of main) I'll put this on CW.

app/app.go
```diff
@@ -352,6 +366,14 @@ func New(
 	return app
 }
 
+func GetDefaultBypassFeeMessages() []string {
+	return []string{
+		sdk.MsgTypeURL(&ibcchanneltypes.MsgRecvPacket{}),
```
giansalex, line 371: this will work considering the gas limitation?
https://github.com/cosmos/gaia/blob/v8.0.0-rc/x/globalfee/ante/fee.go#L67

Tx gas: (>400k)
https://www.mintscan.io/juno/txs/1DD7ADE4C7993554BB13BA83BBCE816287B786BDE784AC42B73F6EC2CF2C5181

Reecepbcups: Probably not, good shout. I'll push a PR upstream to change via ante directly so we don't have to PR. Thanks! Missed this

PR #427, opened 2022-12-17
use mev tendermint by default

faddat:
This PR makes mev-tendermint Juno's default tendermint implementation.

It is being discussed in this gov prop:

* https://commonwealth.im/juno/discussion/8099-promote-healthy-mev-within-the-juno-community-and-set-ecosystem-limits-on-harmful-mev

Once mev-tendermint is Juno's default tendermint consensus implementation, it becomes possible for Juno to work to put features in (or remove features from) the core consensus layer.

A while ago, I experimented with what that might look like:
* #404


faddat:
Due to the reality that the normal maintainer ship of tendermint can no longer maintain that repository because copyright holders have asserted their dominance over that repository, I support merging this PR ASAP.

go.mod
```diff
@@ -165,3 +165,5 @@ replace (
 	// use grpc version that's compatible with cosmos-flavored protocol buffers
 	google.golang.org/grpc => google.golang.org/grpc v1.33.2
 )
+
+replace github.com/tendermint/tendermint => github.com/skip-mev/mev-tendermint v0.34.24-0.20221216212242-89790ca280e5
```
vuong177, line 169:
```suggestion
replace github.com/tendermint/tendermint => github.com/skip-mev/mev-tendermint v0.34.24-mev.13
```
Should we use a tag instead of a commit hash ?

faddat:
I think it makes sense to just close this for now and wait and see if we have any sensible outcome from upstream. Honestly I don't think any outcome here is going to be all that sensible but let's see.

PR #430, opened 2022-12-18
fix logic updateFeeShare

### LINKED ISSUE #429: Logic UpdateFeeShare not correct
Logic UpdateFeeShare not correct

Currently, the UpdateFeeShare not working (bug in logic) https://github.com/CosmosContracts/juno/blob/7c6c26692bc03711382dbac271e9163ceb8ba6a0/x/feeshare/keeper/msg_server.go#L145-L173
What we need here is replace `feeshare.WithdrawerAddress` with `msg.WithdrawerAddress`. But currently, the code get the `withdrawAddr` from `feeshare.WithdrawerAddress` https://github.com/CosmosContracts/juno/blob/7c6c26692bc03711382dbac271e9163ceb8ba6a0/x/feeshare/keeper/msg_server.go#L156
and just re-add the `feeshare.WithdrawerAddress` with the `withdrawAddr` => No changed
https://github.com/CosmosContracts/juno/blob/7c6c26692bc03711382dbac271e9163ceb8ba6a0/x/feeshare/keeper/msg_server.go#L172

GNaD13:
Close #429

### DIFF
--- a/x/feeshare/keeper/msg_server.go
+++ b/x/feeshare/keeper/msg_server.go
@@ -160,16 +160,16 @@ func (k Keeper) UpdateFeeShare(
 			"invalid withdrawer address (%s)", err,
 		)
 	}
+	newWithdrawAddr, err := sdk.AccAddressFromBech32(msg.WithdrawerAddress)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


faddat (approved):
very good find!

Reecepbcups (approved):
Ty, great find

PR #432, opened 2022-12-18
old - packet forward middleware

faddat:
This PR implements the packet forward middleware.

It's using the branch referenced in this pr:

* https://github.com/strangelove-ventures/packet-forward-middleware/pull/45

Since the implementation with ibc v4 is a bit different from v3, reviewers should have a look at:

https://github.com/cosmos/gaia/blob/6880cd7dc22ada30b9dd5737d45f1f90d38a5158/app/keepers/keepers.go#L323

We should also write some e2e tests for this integration, to ensure that it works correctly.


faddat:
This merges the [faddat/make-proto-only](https://github.com/CosmosContracts/juno/tree/faddat/make-proto-only) branch in here, to ensure that it was the needed fix.

faddat:
@vuong177 @Reecepbcups @GNaD13 - guys, getting some pretty serious proto cross-talk here. 

Also over here:

https://github.com/CosmosContracts/token-factory/pull/1

faddat:
@Reecepbcups fyi I think that the proto changes you made to feeshare resolved the issues here.

faddat:
To guide review:

https://github.com/CosmosContracts/juno/actions/runs/3779340500/jobs/6424514084#step:9:13

app/keepers/keepers.go
```diff
@@ -298,8 +312,18 @@ func NewAppKeepers(
 		scopedICAControllerKeeper, bApp.MsgServiceRouter(),
 	)
 
+	appKeepers.ICAModule = ica.NewAppModule(&appKeepers.ICAControllerKeeper, &appKeepers.ICAHostKeeper)
 	icaHostIBCModule := icahost.NewIBCModule(appKeepers.ICAHostKeeper)
 
+	appKeepers.RouterKeeper = routerkeeper.NewKeeper(
```
Reecepbcups, line 318: new keeper, so the key should be added to the v12 upgrade handler store & any params required

Reecepbcups (changes requested):
Also should add an e2e test for this to ensure it works correctly

faddat:
@Reecepbcups I agree about e2e

I also think that this is a reasonable place to segue into possibly using IBC test, maybe after we do the UNI deployment.

But I still do need to figure out how to use it to simulate the test against the live state.

Reecepbcups:
Push the v13 if we can't figure out where test are, which just leaves oracle left for review

faddat:
Yeah I think that makes perfect sense

Reecepbcups:
**Monday Meeting** Pushed to v13 officially

faddat:
Did a merge here for some testing for:

https://github.com/strangelove-ventures/packet-forward-middleware/pull/45

Reecepbcups:
Closing in favor of Andrews more simple PR without go.mod replace (since its merged in now to mainline)
Will add E2E test after that

PR #434, opened 2022-12-19
DNM: incomplete v13 with 46

faddat:
Just in case we need it

...but looks like we won't!


faddat:
I'm going to close this because I think that we can genuinely work with 47 for the next release

PR #435, opened 2022-12-19
use a token factory with ibc-go/v4

faddat:
we had an ibc version mismatch in Juno v12 / token factory.

This PR resolves that :)


Reecepbcups:
@faddat I don't see anything changed here? It is supposed to replace CosmWasm/token-factory to CosmosContracts/token-factory with ibc v4 & wasmd 30 yes?

faddat:
thank you good catch.

I'll get this cleaned up tomorrow.

faddat:
* https://github.com/CosmosContracts/token-factory/pull/1/commits

faddat:
I'd like to get this one in ASAP, same as Oracle, so that we can begin to test against it.  I'm still puzzled at the issues in #432

vuong177 (approved):

Reecepbcups (approved):

PR #440, opened 2022-12-21
Create ossar.yml

faddat:
not sure that this will work because it runs on a windows runner, but basically wanted to test


faddat:
wow, way too "windows"

PR #441, opened 2022-12-21
Create dependency-review.yml

faddat:
to ensure we don't add vulnerable dependencies


Reecepbcups:
Could see an issue where we need to push out a minor patch, but are required to do a full upgrade due to state breaking nature of a dependency forced upgrade. 

Will defer to others for approval on this if that is a non issue

faddat:
Yeah I think that what you mentioned is not really an issue. This basically protects us against supply chain attacks.

Also if you like the skip PR, let's just merge that thing.  upstream pooped in the river.

Reecepbcups (approved):

JakeHartnell (approved):

Issue #443, opened 2022-12-22
Tracking issue for price history features


vuong177:
## Background
This issue outlines the price history feature and QA required for all the changes.

## Work Breakdown
- [x] #444
- [x] #445
- [x] #446
- [x] #447
- [x] #448

GNaD13:
Got it

PR #450, opened 2022-12-24
Dang/add twap

faddat:
Making a PR for dang's twap work for the oracle.


GNaD13:
> Making a PR for dang's twap work for the oracle.

thank you sir :3. Currently, We have some change in the price history logic. The detail is described here #443. The change may affect to TWAP, so I need to work on it first and will back to TWAP asap :3

faddat:
I think that we should try to get the oracle merged, that way these PR's target main

Reecepbcups:
@GNaD13 This can target main now (oracle merged)

GNaD13:
> 

Oh, I forgot this PR, I think this PR can just easily close. Everythings already on Oracle. 

> > Making a PR for dang's twap work for the oracle.
> 
> thank you sir :3. Currently, We have some change in the price history logic. The detail is described here #443. The change may affect to TWAP, so I need to work on it first and will back to TWAP asap :3

As I said here, the new logic have been already implemented in Oracle. So this PR can just close

PR #459, opened 2022-12-27
Add comment and change func name more sense

### LINKED ISSUE #457: Use stargate query for token factory.
## Background

- Wasmd v0.30.0 has stargate query, so we do not need wasmbinding for query chain state from smart contract.
- Notional team are building a rust library for stargate here : https://github.com/notional-labs/juno-rust.

## Suggested Design
- Migrate tokenfactory custom query to stargate query
- Write new contract for tokenfactory testing

## Acceptance Criteria
- Have new contract for e2e test
- All test passed

GNaD13:
Add some comment and change func name for more detail


vuong177:
hey @GNaD13, I think #457 should be solved in a separate PR.

PR #463, opened 2022-12-29
Oracle tests: multiple denom, denom with no entries; endBlocker tests

udit-gulati:
1. Adds price history test for when we have multiple denominations (one denom is prefix of another also covered), to make sure that multiple denom entries does not interfere with each other.
2. Adds price history test for when price is queried for denom which hasn't been stored yet.
3. Adds abci tests for EndBlocker to see if price is correctly stored after validator vote period is over.

To discuss:
1. `getHistoryEntryBetweenTime()` returns `[]types.PriceHistoryEntry(nil)` instead of `[]types.PriceHistoryEntry{}` when the denom does not have stored prices. Is it expected or should be changed?
2. `GetArithmetricTWAP()` throws errror `no values in range` when there is no price entry for given denom before `startTime` becuase the first and last price entry is fetched using `getHistoryEntryAtOrBeforeTime()` and it throws error for `startTime`. IMO this should be handled in `GetArithmetricTWAP()` and values must be returned even if there is no price entry before `startTime`.
3. I've left TODO in comments for both of the above.

### DIFF
--- a/x/oracle/keeper/history_test.go
+++ b/x/oracle/keeper/history_test.go
@@ -154,3 +154,231 @@ func TestRemoveHistoryEntryBeforeTime(t *testing.T) {
 	require.Equal(t, 1, len(phStores))
 	require.Equal(t, phStores[0], phEntrys[2])
 }
+
+func TestStoreAndGetMultipleHistoricalData(t *testing.T) {
+	timeNow := time.Now().UTC()
+	ctx, keepers := CreateTestInput(t, false)
+	oracleKeeper := keepers.OracleKeeper
+
+	phEntrysDenom := []types.PriceHistoryEntry{
+		{
+			Price:           sdk.OneDec(),
+			VotePeriodCount: 10,
+			PriceUpdateTime: timeNow,
+		},
+		{
+			Price:           sdk.OneDec(),
+			VotePeriodCount: 11,
+			PriceUpdateTime: timeNow.Add(time.Minute * 2),
+		},
+		{
+			Price:           sdk.OneDec(),
+			VotePeriodCount: 12,
+			PriceUpdateTime: timeNow.Add(time.Minute * 4),
+		},
+	}
+	for _, phEntry := range phEntrysDenom {
+		oracleKeeper.storeHistoricalData(ctx, "Denom", phEntry)
+	}
+
+	phEntrysDen := []types.PriceHistoryEntry{
+		{
+			Price:           sdk.OneDec(),
+			VotePeriodCount: 100,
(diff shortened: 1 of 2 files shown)


x/oracle/keeper/history_test.go
```diff
@@ -154,3 +154,231 @@ func TestRemoveHistoryEntryBeforeTime(t *testing.T) {
 	require.Equal(t, 1, len(phStores))
 	require.Equal(t, phStores[0], phEntrys[2])
 }
+
+func TestStoreAndGetMultipleHistoricalData(t *testing.T) {
```
vuong177, line 158: This test should be split into smaller tests, it's too long and hard to track.

vuong177:
@udit-gulati thanks for this PR. I think we should use driven-table test cases for this, it's easier for tracking and debugging.

vuong177 (changes requested):
`TestStoreAndGetMultipleHistoricalData` should be split into smaller tests, it's too long and hard to track.

x/oracle/keeper/history_test.go
vuong177, line 158: it should be split into smaller tests.

x/oracle/keeper/history_test.go
```diff
@@ -154,3 +154,231 @@ func TestRemoveHistoryEntryBeforeTime(t *testing.T) {
...
+		phEntrysJuno[1].PriceUpdateTime,
+	)
+	require.NoError(t, err)
+	require.Equal(t, 2, len(phStoresdjuno))
+	require.Equal(t, phStoresdjuno[0], phEntrysJuno[0])
+	require.Equal(t, phStoresdjuno[1], phEntrysJuno[1])
+}
+
+func TestStoreAndGetNullHistoricalData(t *testing.T) {
+	timeNow := time.Now().UTC()
```
vuong177, line 339: Using a driven table test for this case could make more clear && easier to debug!

vuong177:
> 1. getHistoryEntryBetweenTime() returns []types.PriceHistoryEntry(nil) instead of []types.PriceHistoryEntry{} when the denom does not have stored prices. Is it expected or should be changed?

it's nil slice and empty slice, those thing is different but their observable behavior is the same.

faddat:
Very nice. Thank you for your work Udit.

@vuong177 Does this conflict with the table driven testing PR that I just merged?

If it adds additional test coverage, I'm likely in support of it. Thanks for your review.

faddat:
@vuong177 -- checking through this, my feeling is that we should merge, then triage in an additional issue.  I'm happy to have the coverage.

Issue #468, opened 2023-01-02
Dependency review action


faddat:
The dependency review action seems to exhaust its api quota really quickly. 

Should we drop it?

https://github.com/CosmosContracts/juno/actions/runs/3820825923/jobs/6499415063

PR #470, opened 2023-01-02
Delete dependency-review.yml

faddat:
closes: #468 

Let's get rid of this -- I LOVE what it does, but it doesn't do that reliably.


Reecepbcups (approved):
agree

vuong177 (approved):

PR #479, opened 2023-01-08
[v12] Set GlobalFees on Upgrade

Reecepbcups:
- Sets to default, was tricky since it has to run after the migrations
- Sending IBC packets is now feeless < 1_000_000 gas as originally intended

0.0025ujuno,0.001ATOM (IBC denom given in docs
`junod q ibc-transfer denom-trace ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9`

```sh
#after upgrade shown successful

junod q globalfee minimum-gas-prices --node http://localhost:26657

minimum_gas_prices:
- amount: "0.002500000000000000"
  denom: ujuno
- amount: "0.001000000000000000"
  denom: ibc/C4CFF46FD6DE35CA4CF4CE031E643C8FDC9BA4B99AE598E9B0ED98FE3A2319F9
```


vuong177:
https://github.com/notional-labs/juno-rust

faddat (approved):

app/upgrades/v12/upgrades.go
```diff
@@ -58,9 +65,11 @@ func CreateV12UpgradeHandler(
 			AllowedDenoms:   []string{nativeDenom},
 		}
 		keepers.FeeShareKeeper.SetParams(ctx, newFeeShareParams)
+		logger.Info("set feeshare params")
 
 		// ICA - https://github.com/CosmosContracts/juno/blob/integrate_ica_changes/app/app.go#L846-L885
 		vm[icatypes.ModuleName] = mm.Modules[icatypes.ModuleName].ConsensusVersion()
+		logger.Info("upgraded icatypes version")
```
dimiandre, line 72: do we need all these logs?

dimiandre (changes requested):

app/upgrades/v12/upgrades.go
faddat: I'm going to say yes.

It'll only happen once but will be nice to know.

dimiandre: ok

dimiandre (approved):

app/upgrades/v12/upgrades.go
```diff
@@ -119,10 +128,29 @@ func CreateV12UpgradeHandler(
...
 		icamodule.InitModule(ctx, controllerParams, hostParams)
+		logger.Info("upgraded ica module")
 
 		// IBCFee
 		vm[ibcfeetypes.ModuleName] = mm.Modules[ibcfeetypes.ModuleName].ConsensusVersion()
+		logger.Info(fmt.Sprintf("ibcfee module version %s set", fmt.Sprint(vm[ibcfeetypes.ModuleName])))
+
+		// Run migrations
+		versionMap, err := mm.RunMigrations(ctx, cfg, vm)
+
```
vuong177, line 139: we should handle err here.

Reecepbcups: We return the err and It’s handled up the stack. This is done just so we can run the globalfee after. (Since before we just did "return RunMigrations()"

PR #490, opened 2023-01-11
[v12] Oracle Setup Documentation

Reecepbcups:
Authored by Oni-Validator (Vince / kw1k)

Tested on my fresh ubuntu machine


faddat:
I am going to read, and most likely approve.  I'm also creating an issue, to ensure that we go back through the docs, before mainnet oracle use, so no one 🔪

faddat (approved):
Approving, with the assumption that this is missing something important and we learn what that is in the Uni testnet.

x/oracle/docs/README.md
```diff
@@ -0,0 +1,259 @@
...
+# If you get `cp: cannot create regular file '$HOME/.juno/oracle-config.toml': No such file or directory`, 
+# run: `junod init [moniker]`. Then run the cp command again
+
+price-feeder version # sdk: v0.45.11 and go1.19.*
+# If it is not found, your go path is not set
+```
+
+---
+
+## Screen
```
JakeHartnell, line 51: Feel like this should go below `systemd` as I think that's a more robust way of doing it? Do any validators actually use `screen`?

JakeHartnell (approved):
Minor comment, otherwise LGTM.

x/oracle/docs/README.md
Reecepbcups: I agree systemd is better, but screen is the more beginner friendly way for most (which is why we did not include tmux). This is why I had it on top

Feel most will be between systemd and the docker

PR #491, opened 2023-01-11
[v12] Oracle - remove redundent AcceptList

Reecepbcups:
Note to myself for Thrus: 
- rename to Whitelist?

From my review - AcceptList seems redundant, and PriceTrackingList does exactly the same thing. 

**Question**: 
Why do AcceptList and PriceTrackingList keys both point to DefaultAcceptList for genesis. 
Is there a reason this is redundant?

Would these 2 ever diverge? 
Why would we accept a denom which we don't want to track? 
in which case would accepted denoms not be tracked?


vuong177:
@Reecepbcups `accept list` is the list of tokens that have oracle in Juno. `price_tracking_list` is the list of tokens with price data history and TWAP. It's different.

Reecepbcups:
@vuong177 when would we accept a token in the oracle which is not tracked by TWAP?

Reecepbcups:
> TWAP requires saving the price history in the chain state; we need to create a list of tracked tokens.

The list of tracked tokens IS what we want to have in the TWAP though yes? In which case AcceptList is redunant.

What this PR does is merges the 2, so any we accept ALSO are the TWAP. So JUNO, ATOM, and OSMO are all accepted & TWAP. If we remove OSMO, then it no longer does TWAP either

faddat:
I am honestly unsure.

I think that what @vuong177 is saying is that when we add a token to twap, it will be saved in state, not HistoricalData.  

Could be wrong, so I will wait to approve.

vuong177:
why we need to combine 2 seperate things to one ? @Reecepbcups

vuong177:
More clearly if we use 2 seperate params. And because the price data save in state, we should limit the number of tracking history token. Also when have stable coin like USDT or USDC, save history price of those token is unnecessary.

vuong177:
We can change multiple parameters in one prop, right? If those parameters for different use cases we should separate it.

Reecepbcups:
Alright sounds good, will keep it how it is then thx

vuong177:
seems`AccpectList` and `PriceTrackingList ` name make confuse, we should have better variable names.

PR #510, opened 2023-01-14
use informal tendermint with sadness

faddat:
This PR ensures that Juno is safe from the odd behavior occuring at 
tendermint/tendermint.


JakeHartnell (approved):
:smiling_face_with_tear:

Reecepbcups (approved):

PR #511, opened 2023-01-14
[v12] Update Swagger API endpoints

Reecepbcups:
This took forever, but finally got a working script to regen endpoints for the openapi 1317 endpoint. 

Needed since we have custom Juno modules now
also maps juno's mint module correctly (was still using cosmos before)

image


the-frey (approved):

faddat:
very good

faddat (approved):

PR #512, opened 2023-01-15
feat(scripts) - Relay CW20 between localnets example

Reecepbcups:
Relays a CW20 with the CW20_ICS20 contract. Tested locally, and tested with mainnet no issues (hermes)

image


the-frey:
This looks really good

the-frey:
My additional thought is that @Reecepbcups has contributed so many useful bash scripts, at some point we should catalogue them better. Whether that's in their own repo or in some way within this one so they're clearly identified &/or thus re-usable to even other chains, since it's all good work

the-frey:
Sounds great 👍

faddat:
Reece, thanks a ton for checking into this in detail. 

Generally, still happy about monorepos, and prefer things to target this repository

faddat:
Just so you know I think this actually should be merged. It's really valuable documentation and thank you for making it.

Reecepbcups:
@faddat I agree, I just need to spend time to clean it up & add CW721 examples, Ill get back to it this weekend most likely

faddat (approved):
very approve, and very like that it is here in the main repository directly.

JakeHartnell (approved):

Issue #515, opened 2023-01-16
Workflow build.yaml is not run 


anhductn2001:
## Background

> Workflow build.yaml do not run because `[Error: .github#L1] every step must define a `uses` or `run` key`

## Acceptance Criteria

> Make this workflow run normally

Reecepbcups:
Fixed in #509 + test in #514 @anhductn2001

PR #516, opened 2023-01-16
Fix build.yml

anhductn2001:
Close issue #515


Reecepbcups:
This was fixed in 511

PR #517, opened 2023-01-16
Update test for TWAP when startTime before first entry

udit-gulati:
TWAP now returns a value in place of error when start time is before first entry, so now test reflects that


Reecepbcups (approved):
Thanks! Yes this should be the expected behavior. Easier for people using TWAP and trying to over extend timing

faddat (approved):

PR #525, opened 2023-01-21
Fix v12 Upgrade Handler

Reecepbcups:
ICA module InitModule issue

This was originally done in v8.0.0 upgrade [v8 upgrade handler](https://github.com/CosmosContracts/juno/blob/v8.0.0/app/app.go#L835-L841) 

which originally added the controller & host, controller was just disabled (I thought it had not been added at all)

All we needed to do was update the params of both (enable controller, add more Msgs to Host)

---
roll back to block XXX500
fix ICA upgrade

if doesnt work, remove GlobalFee?

if doesnt work, FeeShare / Oracle?

### DIFF
--- a/app/upgrades/v12/upgrades.go
+++ b/app/upgrades/v12/upgrades.go
@@ -11,8 +11,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/authz"
 
 	// ICA
-	ica "github.com/cosmos/ibc-go/v4/modules/apps/27-interchain-accounts"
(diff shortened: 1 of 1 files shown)


faddat (approved):
Looks good.  Uni has gas that isn't level, however.  

I like your suggestion about reverting to a pre-upgrade state and running the upgrade again. If we can get all the validators in one place then it probably makes sense to try it with this before investigating gas

Reecepbcups:
✅ https://quicksync.cros-nest.com/transfer/uni-5-1785500.tar.lz4

JakeHartnell (approved):

PR #527, opened 2023-01-23
[feeshare] Call fee check AFTER there is someone who must be paid

Reecepbcups:
title - don't sort fees if we dont need too


JakeHartnell (approved):
LGTM

faddat:
Merging in man and having a peek

faddat (approved):
Looks quite sensible from where I'm standing

PR #530, opened 2023-01-24
DNM - Ez Multi-Remote testnet generation

Reecepbcups:
using to test v11 -> v12 upgrade with multiple nodes semi automatically. Is decent


Reecepbcups:
Keeping Local

PR #533, opened 2023-01-25
chore: use upstream updates

faddat:
* sdk v0.45.12
* ibc-go v4.3.0
* remove dragonberry replace because we're using ics23 v0.9.0


Reecepbcups:
Breaks IAVL tree every 15 blocks on a fresh instance after upgrade. Not production ready

PR #536, opened 2023-01-26
remove globalfee

Reecepbcups:
removes globalee
regenerates swagger docs

does not touch anything else as v45.12 and ibc 4.3.0 break every 15 blocks for the IAVL tree it seems


JakeHartnell (approved):
RIP

faddat (approved):
It's too bad. I liked this one.

PR #538, opened 2023-01-30
add Unix time format to twap query

vuong177:
- Add UNIX format to GetCmdQueryTwapPrice


Reecepbcups (approved):
utAck

Reecepbcups:
The team has decided not go with this oracle impl. Due to the non deterministic nature during Uni-5

PR #544, opened 2023-02-01
Newer version of IBC hooks that always succeeds on timeouts

nicolaslara:
IBC hooks allows for a sender to call a callback when the IBC lifecycle finishes. If that contract fails or doesn't exist on a timeout, the previous version of IBC hooks would error out preventing the timeout to happen. 

This version always allows timeouts to complete and emits the error instead. 

I think it would make sense to include this update in v13


JakeHartnell (approved):

the-frey:
Makes sense to include, yep. What's the CI failure? Looks unrelated to this PR

nicolaslara:
yeah, no idea what the CI failure is. I noticed that main also was failing and that that particular check didn't execute on main, so I didn't dig into it.

Reecepbcups (approved):
utACK, Yea we can fix lint issues later. CI is all over the place atm it seems

PR #551, opened 2023-02-07
Bump forward middleware to v4.0.4

agouin:
https://github.com/strangelove-ventures/packet-forward-middleware/releases/tag/v4.0.4

Resolves an issue with JSON string encoded `next` metadata when there is more than one hop. JSON object encoded `next` metadata was unaffected.


faddat:
Thank you so much

PR #558, opened 2023-02-11
go workspaces

faddat:
A recent Juno dev meme has been approximately "let's tidy the price feeder!"

But what if we could simply "go work sync" and it would be tidied for us?

Dream no more, go workspaces are here :)


JakeHartnell (approved):
Oh nice, didn't know about this feature.

faddat:
This is kinda waiting on these:

* https://github.com/strangelove-ventures/packet-forward-middleware/pull/51
* https://github.com/strangelove-ventures/packet-forward-middleware/pull/70

Reecepbcups:
I'll approve when this works, though I did add "make all" to go mod tidy and lint both

I guess this is the more correct way to do that

faddat:
Yeah it is waiting on some upstream stuff.

PR #560, opened 2023-02-14
v12: bump tendermint, remove ignite, sdk, ibc

faddat:
This PR:

* updates ibc-go to v3.4.0
* updates tendermint to use the informal fork, v0.34.26
* updates cosmos-sdk to v0.45.13
* moves the openapiconsole so that we no longer pull the ignite cli into dependencies


Reecepbcups:
Not tested on Uni, so going to have to say no to this one. Will be apart of v13 already though

Reecepbcups:
I do not really see a reason for this in v12 since we are moving to v13 so quickly which will have these changes & no way to test easily with Uni since its on v13. Closing

PR #571, opened 2023-02-23
DNM: feat(wasm): Gas limit on execute to increase block gas param

Reecepbcups:
**CI Failure is JunoSwap API down for price feeder**

**NOTE: NOT SURE IF THIS ACTUALLY SOLVES THE ISSUE YET OR NOT, I CAN NOT REPLICATE YET**

TODO: 
- backport to v13?

---

With v13 we need to increase gas limits. This allow us to increase block limits while limiting wasm execute calls max gas per block via ante.

It can easily be added to do the same for a migration & instantiate (should put at a higher limit?)

So we can increase block gas back to 100mil, then limit an execute to like 15-25m per transaction?


JakeHartnell (approved):

faddat:
Hey, I need to poke this a bunch, that is what youd like?

Reecepbcups:
@faddat we need to ensure the current testnet is good. If it is, then there is no reason to add this. However if we can crash uni, then we need to make a release for this patch if it works

I'll run the weapon's lab code on Uni in the coming days to test it.

Reecepbcups:
not needed due to 100m gas being good to go now 
https://commonwealth.im/juno/discussion/10396-increase-maximum-per-block-gas-to-100-million

PR #577, opened 2023-02-28
use comet

faddat:
This PR replaces tendermint with cometbft v0.34.27, and bumps cosmos-sdk 
to v0.34.15-rc.0

It shouldn't be shipped to Uni for the upcoming v13 ugrade, but instead 
should be a part of v14.


Reecepbcups:
Please put this in #576 since it is for v14+ (main branch) so all bumps are in a single PR. Thanks

PR #578, opened 2023-02-28
packet forward middleware 405

faddat:
This PR bumps packet forward middleware to v4.0.5 and likely should be 
shipped as part of v13.


Reecepbcups:
This is already done in #576, we are going to keep bumps in a single PR for main going forward as previously discussed.

It is also already running with v13.0.0-beta.2 on Uni-6 testnet :)

PR #588, opened 2023-03-10
[main] Version bumps working branch

Reecepbcups:
Any version bumps (excluding the skip protocol change) commit to this branch

Update: that are non state breaking


faddat:
I think this idea of a "version bumps branch" is harmful to the idea of keeping things up to date consistently.

faddat:
I think that the new WASMD changes the commands for deploy maybe, which might be causing the failures here.  Will poke this tomorrow.

Reecepbcups:
> I think this idea of a "version bumps branch" is harmful to the idea of keeping things up to date consistently.
> 
> Recommend that we get this merged, so that we know what we are building against.

This is just for version bumps which are non critical (ex: PFM from 4.0.3 to 4.0.4 for example). If something is critical and needed for another PR, it can be done in other PRs. Having a PR for every minor version bump or change or go version difference is inefficient to constantly review, merge approve and repeat.

Happy for wasm .31 to be in its own PR if its more work than a version bump

faddat:
So I remain in disagreement mainly because sometimes even minor patches are very important, but for now lemme see what I can do about 31.

faddat:
It turned out being very much an issue of making and testing changes one by one instead of in one big branch:

https://github.com/CosmosContracts/juno/pull/602

faddat:
suggest we close this

Reecepbcups:
I agree, that merge & trying to put state breaking changes in here did not work. Future: only non state breaking changes go in here for very simple version bumps (so we don't have to approve / merge 30+ PRs per review cycle for non production changes)

PR #601, opened 2023-03-18
feat(cli): Specify client.toml variables (gas, gas-prices, note, etc)

### LINKED ISSUE #518: Reducing flags in `junod` commands
## Background

Juno devs use `junod` a lot, and there are a series of flags we have to copy/paste that could potentially be reduced. It would be great if we could simplify this:

```
junod tx wasm execute juno14q8kk464fafql2fwmlsgvgcdl6h2csqpzv4hr025fmcvgjahpess32k0j7 \
'{"burn":{"amount":"42"}}' --node https://juno-rpc.polkachu.com:443 --chain-id juno-1 \
--gas-prices 0.025ujuno --gas auto --gas-adjustment 1.3 --from myaccount -y
```

to

```
junod tx wasm execute juno14q8kk464fafql2fwmlsgvgcdl6h2csqpzv4hr025fmcvgjahpess32k0j7 \
'{"burn":{"amount":"42"}}' --from myaccount -y
```

Some of these flags are configurable, which is great, but not all of them.

If you run `junod config` you can see something like:

```json
{
  "chain-id": "juno-1",
  "keyring-backend": "os",
  "output": "json",
  "node": "https://juno-rpc.polkachu.com:443",
  "broadcast-mode": "block"
}
```

By default, it seems these are stored in `~/.juno/config/client.toml`, but you cannot add entries (manually or with the `junod config` command) for a few flags, namely:

- `gas-prices`
- `gas`
- `gas-adjustment`

Since those values cannot be loaded from the TOML config, they must be added manually. It seems that it's in the `cosmos-sdk` repo. 

You can see one such flag being set here:
https://github.com/cosmos/cosmos-sdk/blob/7505c18e2f8f695dd6ab7b00515a9686bb5cec27/client/flags/flags.go#L18

which, for Juno, would be `1.3`

and then farther down, it seems there isn't a default given here, but I believe we could:
https://github.com/cosmos/cosmos-sdk/blob/7505c18e2f8f695dd6ab7b00515a9686bb5cec27/client/flags/flags.go#L117

So this could have a default of `0.025ujuno` for mainnet and `0.025ujunox` for testnet

From what I've heard, and then learned the hard way last night, it seems that forking `cosmos-sdk` and modifying my local `juno` to that (ie, `go get github.com/mikedotexe/cosmos-sdk/client/flags@v0.45.11-juno`) will not work because of how imports work in Go; or something like that.

You chase errors like this:

```
../../go/pkg/mod/github.com/cosmos/ibc-go/v3@v3.3.1/modules/core/02-client/types/client.pb.go:9:2: github.com/mikedotexe/cosmos-sdk@v0.45.11: parsing go.mod:
	module declares its path as: github.com/cosmos/cosmos-sdk
	        but was required as: github.com/mikedotexe/cosmos-sdk
```

So how can we remedy this for builders who use dozens of `junod` commands daily?

## Suggested Design

Feel free to correct/brainstorm, but I believe we might be able to have chain-specific defaults if we use git tags on the official `cosmos/cosmos-sdk` repo.

Note the require here:
https://github.com/CosmosContracts/juno/blob/ef0e81c79cc076c89221f8c109cf63d4ddfcd0eb/go.mod#L7

Instead, what if this linked to a tag specifically for juno mainnet or juno testnet. (Two separate ones because default price would be "ujuno" vs. "ujunox")

So the require would instead be:

    github.com/mikedotexe/cosmos-sdk v0.45.11-chain-juno
    or
    github.com/mikedotexe/cosmos-sdk v0.45.11-chain-junotestnet

Where the `chain_name` is used, as it is according to chain-registry, like:

https://github.com/cosmos/chain-registry/blob/ad9cf415745c8864f1d2a91cef5f7cb9b4bf2a6c/testnets/junotestnet/chain.json#L3

------

If we had tags like this, then the community could clone this repo, change 1 line, `make build` and have a `junod` binary specific to that network and reduce the repetitive copy/paste of flags.

It would still be a little clunky, because then you have two binaries, but we could easily use aliases to switch between them.

For instance, an entry in `~/.profile`/`~/.bashrc`/`~/.zshrc`:

```bash
# Switch to Juno mainnet
alias jm='junod config chain-id juno-1 && junod config node "https://juno-rpc.polkachu.com:443" && ln -s /Users/red-team/junod-mainnet /Users/red-team/go/bin/junod'

# Switch to Juno testnet
alias jt='junod config chain-id uni-5 && junod config node "https://rpc.uni.junonetwork.io:443" && ln -s /Users/red-team/junod-testnet /Users/red-team/go/bin/junod'
```

Then you could run `jm` and the config values we *can* change in `client.toml` would change, and a symlink would get updated such that `junod` pointing to the mainnet-specific binary. Same idea for `jt` changing flags for testnet.

## Acceptance Criteria

As a Red team builder, I would like to run all the commands I run now with `junod`, except without having to copy/paste the gas flags every time.

Reecepbcups:
Closes #518 

# CI Failure note
The linter upgraded overnight... 95% of the issues are in the oracle and price feeder - which are being removed anyway in another PR

# Feature
Reads from the client.toml file to get custom defined variables. This includes:
- gas
- gas-price
- gas-adjustment
- fees
- fee-account
- note (memo)

## Logic
- When set in the config, it auto populates by setting the env of said command for that transaction.
-  If a user has a local env set, it will override the config
- `junod config gas 1000000` to add in all the custom configuration options
- If you set gas-prices, it will clear fees. If you set fees, clears gas-prices (since you can not do both)

---

image

## Flow
- Set `junod config gas 1000000` into config.toml
- Run `junod tx bank send key juno1... 100ujuno`
- Program check config, and sets JUNOD_GAS=1000000 for that run
- User then runs `export JUNOD_GAS=500000` in the terminal
- Run `junod tx bank send key juno1... 100ujuno`
- Program sees gas is 500k, and uses that for this run. Ignoring config. (then to use config again, set JUNOD_GAS="")


Reecepbcups:
UGH the linter changed overnight... 95% of the issues are in the oracle and price feeder - which are being removed anyway

faddat:
K, you have a PR to take out the price feeder, yes?

faddat:
PS:  this looks like solid usability improvements

faddat (approved):

Reecepbcups:
> K, you have a PR to take out the price feeder, yes?

@faddat yes, that PR which was just merged to main did this. So I'll pull main in later today & will fix these PRs CI

faddat:
Cherry picked the lint commits from another branch, this should be good to go now, linter-wise.

JakeHartnell (approved):

PR #602, opened 2023-03-19
wasmd 31 & fix interchaintest

faddat:
## Interchain test

* https://github.com/strangelove-ventures/interchaintest/pull/448

For now I'll get it tuned up by using a replace statement, and we can get rid of the replace statement when interchaintest v4 has been updated.  

The basic summary of the issue is that interchaintest was using a version of grpc that isn't compatible with the regen modified protobuf compiler.

Interchaintest also needed to be updated to support wasmd v0.31.0

## wasmd 31

The Dockerfile needed to be updated to be compatible with wasmvm v1.2.1

```Dockerfile
# See https://github.com/CosmWasm/wasmvm/releases
ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.2.1/libwasmvm_muslc.aarch64.a /lib/libwasmvm_muslc.aarch64.a
ADD https://github.com/CosmWasm/wasmvm/releases/download/v1.2.1/libwasmvm_muslc.x86_64.a /lib/libwasmvm_muslc.x86_64.a
RUN sha256sum /lib/libwasmvm_muslc.aarch64.a | grep 86bc5fdc0f01201481c36e17cd3dfed6e9650d22e1c5c8983a5b78c231789ee0
RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep a00700aa19f5bfe0f46290ddf69bf51eb03a6dfcd88b905e1081af2e42dbbafc
```


faddat:
@Reecepbcups @vuong177 -- doesn't this look oddly similar to some of the issues that we had with the oracle?

faddat:
Also noticing that maybe I need to be registered on heighliner?

Reecepbcups:
Would rather us not hard code it - see commit 1c00723b61089de0f3786f24bfabe70885afa892 dockerfile line 20-30 (So it will auto update based off our go.mod)

Reecepbcups:
@faddat The gov contracts failing is due to the unity-contract needing to be updated afaik - though the version issues may be another storeKey issue? was icacontrollertypes last time

faddat:
honestly I am more content with a dockerfie that doesn't update itself

PR #603, opened 2023-03-19
Apply revive more strictly

faddat:
This PR applies the revive linter more strictly.  It is definitely a work in progress.


app/decorators/min_commission.go
```diff
@@ -51,37 +62,13 @@ func (min MinCommissionDecorator) AnteHandle(
 		return nil
 	}
 
-	validAuthz := func(execMsg *authz.MsgExec) error {
```
Reecepbcups, line 54: ? - This is needed to ensure a validator does not do an Authz message to bypass the minimum commission amount

Or else people can bypass with 0 commission through an authz create new validator msg

faddat: 100%.  Thank you.

faddat: will revert this file for now.  When I do that I'll let ya know why that happened.

PR #614, opened 2023-03-21
chore(ci): optimize depending on changed files

Reecepbcups:
## Feature
Cleans up CI so its more specific depending on changes. 

## Changelog
- a **.md only change requires no CI
- If we alter **.go, go.mod, or go.sum -> we run all test
- We no longer build a docker image every time we push or PR, since ictest does this now
- If we alter .sh, .bash, or any file with docker in it, we test build docker specifically.


faddat:
This is the first time I have ever liked a PR that uses conditionals in CI. 

Nice.

faddat (approved):
This is great, and native.

JakeHartnell (approved):

PR #619, opened 2023-03-22
[v14]: SDK v45.15 + CometBFT

Reecepbcups:
Once Skip has a CometBFT fork we can switch to that. Non state breaking so we can get in after we launch on Uni if needed

This is the last SDK v45 release. **If users have issues with state sync, they MUST be on go.1.19+ to fix it**


JakeHartnell (approved):
LFG!

faddat:
mutual LFG!

faddat (approved):

PR #623, opened 2023-03-23
update to wasmd v0.32

faddat:
This PR updates Juno to wasmd v0.32.0

(they've yet to tag but are currently just doing release notes:

* https://github.com/CosmWasm/wasmd/pull/1285
)


Reecepbcups:
If this isn't tagged officially by Monday, we will continue on with just 31. Then can bump for 14.0.1

Reecepbcups:
Still not tagged, pushing to v15.

PR #624, opened 2023-03-26
DNM - ictest: Fix local upgrade, packet-forward

Reecepbcups:
## Changes
- Fixes local upgrade test

## ToDo
- Upstream: param-change proposal from io read file to v4
- Add IBC param change test (from 28 days to 14 days) and test IBC transfer after. Breaks?
- Add Akash dev net here off main tip
- On main branch push, run a 20+ validator testnet? How does github handle this?


Reecepbcups:
Ill do later

PR #625, opened 2023-03-28
short block times

faddat:
This PR will shorten Juno's block times. 

With this PR we should get block times of about three seconds and validators should not be able to adjust the commit timeout.


Reecepbcups:
SDK v45 does not have this feature, only v46+ it seems to modify the tendermint config. Closing this PR and opening a patch for this into my v15 initial branch

PR #626, opened 2023-03-29
[v14] Cleanup + Enable CosmWasm 1.2 feature

Reecepbcups:
## Fixes
- Added cosmwasm_1_2 type to whitelist 
- CometBFT from Skip
- Fix ictest: TestBasicJunoStart
- GlobalFee Bypass: MsgTimeout & MsgTimeoutOnClose (Co author @giansalex. Will credit in the release notes)

## Cleanup
- Cancel ictest on new commit if still running in previous
- Moves GetChainsDenomToken to a single function definition
- Bump protobuf version


JakeHartnell (approved):
LGTM

Reecepbcups:
Hmm main upgrade keeps failing even though I changed nothing in the normal v14 here.

Thoughts:
- Running 8 val nodes + 8 full nodes for testing kills GI CI to start the new ones (from v13->v14)
- gogo proto bump broke things
- no clue

Reecepbcups:
Will discuss this Monday, but we can't have 2 upgrade handlers set. only the last one applies even though the upgradeKeeper accepts a map of upgrade names

I wanted to do it safe, but since it is non-state breaking, we can just ask vals to halt at some height, then apply.

---
Reason: 

If v14.Upgrade is last in the upgrade array, then it applies even if we call another upgrades name (univ14part2 in this case)

If univ14part2 is last, then interchain upgrade test fails because it was doing v13 -> this upgrade, even though we tell it "v14"

faddat:
definitely would like to go over this on the call.

Thanks for this @Reecepbcups

Reecepbcups:
I will tag off this as alpha.2 once CI completes. Running on my Uni validator now

.github/workflows/interchaintest-E2E.yaml
```diff
@@ -14,6 +14,10 @@ env:
   REGISTRY: ghcr.io  
   IMAGE_NAME: ${{ github.repository }}-e2e
 
+concurrency:
```
faddat, line 17: nice

faddat (approved):

faddat:
good stuff

PR #627, opened 2023-03-29
include mev-cometbft instead of tendermint

nivasan1:
## In This PR
 - I added a replace statement to replace vanilla cometbft with mev-cometbft 💯


Reecepbcups:
:fire: :fire:

PR #635, opened 2023-04-06
[v15] TokenFactory DenomCreation: GasConsume

Reecepbcups:
## Changes
- DenomCreationGasConsume is set at `2_000_000` gas connsume on create IF/when a param change proposal removes the fee
- Uses the new CosmosTokenFactory repo (Will be a shared repo with other chains) so we all have like bindings

## TODO:
- Add a way to disable sending of a token via bankKeeper. Only allow minting (up to the admin of each coin)


scripts/test_node.sh
```diff
@@ -73,7 +73,9 @@ from_scratch () {
   update_test_genesis '.app_state["gov"]["deposit_params"]["min_deposit"]=[{"denom": "ujuno","amount": "1000000"}]'
   update_test_genesis '.app_state["crisis"]["constant_fee"]={"denom": "ujuno","amount": "1000"}'  
 
-  update_test_genesis '.app_state["tokenfactory"]["params"]["denom_creation_fee"]=[{"denom":"ujuno","amount":"100"}]'
+  # update_test_genesis '.app_state["tokenfactory"]["params"]["denom_creation_fee"]=[{"denom":"ujuno","amount":"100"}]'
+  update_test_genesis '.app_state["tokenfactory"]["params"]["denom_creation_fee"]=[]'
+  update_test_genesis '.app_state["tokenfactory"]["params"]["denom_creation_gas_consume"]=2000000'
```
faddat, line 78: basically, we will make this a high gas tx, instead of a fee'd tx?

Reecepbcups: exactly

This passes the cost on to the user executing, rather than the contract developer directly (if a new denom is needed to be created)

Current mainnet is still 1JUNO cost until larry puts up a para change proposal - [https://commonwealth.im/juno/discussion/10596-parameter-change-reduce-denom_creation_fee-to-zero](https://commonwealth.im/juno/discussion/10596-parameter-change-reduce-denom_creation_fee-to-zero)

larry0x:
@Reecepbcups I see you have changed the protobuf type URLs in the new tokenfactory repo (from `osmosis.tokenfactory` to `cosmos.tokenfactory`)

This can be problematic because many contracts (such as mine) use the [osmosis-std library](https://crates.io/crates/osmosis-std) for interaction with tokenfactory module, and that library uses the `osmosis.*` type URL.

Ideally, please don't introduce breaking changes to the protobuf API 🙏

Reecepbcups:
@larry0x Oh, I changed this due to me working on getting https://github.com/CosmosTokenFactory/token-factory live. Where Juno, Sei, and Kujira will all use the new bindings once I add some test & reach out with them more on it

(Like you discussed you wanted for ics-999).

If this is not ideal I am happy to revert proto back to osmosis.* paths, but figured since its deviating away from that (then they can catch up) this is the best? no?

larry0x:
@Reecepbcups Osmosis team's plan is to spinoff the module into a standalone repo (https://twitter.com/sunnya97/status/1641117493409333252). Not sure how soon this will happen (I will push for it) but once it happens, most chains will most likely import from it. The repo you made will be an interim solution and IMO it should remain compatible with the Osmosis one.

Reecepbcups:
@larry0x Got it thanks. I'll revert proto back then, missed this reply.

faddat:
I think that this should target main instead of a long-hanging branch.

This is a small thing and nbd.

so let's roll.

PR #636, opened 2023-04-08
bump: gaia

faddat:
This PR only bumps cosmos/gaia


Reecepbcups:
I already handle this in v15 here https://github.com/CosmosContracts/juno/pull/635 - we will not have time to get in a v14.0.1 on Uni since I have already almost finished v15.

PR #637, opened 2023-04-08
[v15] Forced Faster Block Times (3 seconds)

Reecepbcups:
**Merges into v15-initial branch**

## Changes
- Pulls out the `server.InterceptConfigsPreRunHandler` function into its own file
- Adds `blockTimeout` and force set it with `config.Consensus.TimeoutCommit` any time the junod command is run

### Bonus Addition
- Adds the wasmconfig to increase gas queries to the config for all

## Check
```bash
CHAIN_ID="local-1" HOME_DIR="~/.juno1" TIMEOUT_COMMIT="100ms" CLEAN=true sh scripts/test_node.sh
```

Even with 100ms set as your config.toml, it still uses the set 3 seconds.

---

## Reason
SDK v46/47 fix this by allowing direct modification of the tendermint config. This is why it has to be "soft forked" into a single file. 

Other chains which have implemented this feature already hard forked the SDK. Would rather not

### DIFF
--- a/cmd/junod/cmd/root.go
+++ b/cmd/junod/cmd/root.go
@@ -5,9 +5,11 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/client"
+	serverconfig "github.com/cosmos/cosmos-sdk/server/config"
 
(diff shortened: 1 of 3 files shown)


faddat:
> Would rather not

With you here.  I think there's another way and I've forgotten how.  For now, I am in approval sir.

faddat (approved):

PR #638, opened 2023-04-10
Fix docker image creation for mainnet releases

Reecepbcups:
interchain test now does this every push, we do not need for tag since Osmosis e2e was removed


faddat (approved):

faddat:
sounds like good problems, thanks!

JakeHartnell (approved):

PR #640, opened 2023-04-11
v16: async-icq

faddat:
This PR was made by referencing Strangelove's async-icq simapp and Osmosis' querier configuration.

I think that this should suit our purposes fine, and give us fully async-icq support.


Reecepbcups:
This does not build & we can get it into the SDK v47 upgrade whenever that comes. I fixed it there

in the meantime Polytone from Ekez will do this - so adding is not a priority

PR #642, opened 2023-04-18
chore: bump wasmvm

the-frey:
Updating wasmvm to the latest version. Child's stuff.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4 // indirect
 	github.com/99designs/keyring v1.2.1 // indirect
 	github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d // indirect
-	github.com/CosmWasm/wasmvm v1.2.1 // indirect
+	github.com/CosmWasm/wasmvm v1.2.3 // indirect
 	github.com/DataDog/zstd v1.5.0 // indirect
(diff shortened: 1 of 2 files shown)


Reecepbcups (changes requested):
Actually we do not need to do a replacement since we already use the wasmvm

Reecepbcups (approved):

the-frey:
Good spot.

dimiandre (approved):
lgtm

JakeHartnell (approved):

go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4 // indirect
 	github.com/99designs/keyring v1.2.1 // indirect
 	github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d // indirect
-	github.com/CosmWasm/wasmvm v1.2.1 // indirect
+	github.com/CosmWasm/wasmvm v1.2.3 // indirect
```
webmaster128, line 37: Is this safe to never revert back? I mean all your direct dependencies only ask for 1.2.1 (`github.com/CosmWasm/wasmd v0.31.0`). But I'm not an expert wrt. Go build systems and `indirect` is a mystery to me.

webmaster128 (approved):

go.mod
robert-zaremba: In case of doubt, always check `go.sum` - this is what is being used.

PR #658, opened 2023-04-30
Add allowed Stargate query: distribution `DelegationRewards`

mikedotexe:
I believe it will be beneficial to Juno to add another allowed Stargate query.

See this query that we're talking about:
https://github.com/cosmos/cosmos-sdk/blob/v0.45.15/proto/cosmos/distribution/v1beta1/query.proto#L37-L41

With this addition, we'll see more programmatic claiming and staking of native tokens. Actually, this Twitter thread details how we can actually just delegate instead of the two-step claim-and-delegate flow we're accustomed to:
https://twitter.com/larry0x/status/1621959355221348352

I believe exposing this query doesn't introduce security concerns, and will help delegation UX if we do.

## An example

As an example, check this out… Here's my account with native balance and some tokens that need to be claimed.

image

Then (without claiming) I sent a `delegate` command for 31 JUNO, which can only succeed if both values are combined.

`junod tx staking delegate junovaloper1gp957czryfgyvxwn3tfnyy2f0t9g2p4pvzc6k3 31000000ujuno --from mikereg -y | jq`

https://www.mintscan.io/juno/txs/454F89DDB856E5A7C68E91D7D72884C243012F1FB46B440ECEFDE234342E2EBF

I was expecting that we'd be able to ascertain how many tokens are available to claim using Stargate queries under `cosmos.staking.…`, which are on the safe Stargate list, but the only query I'm seeing that can get this is the one I've added in this PR from `cosmos.distribution.…`.

If we add this Stargate query, then CosmWasm contracts can leverage authz, and efficiently redelegate in a single transaction instead of claiming and delegating, which isn't necessary. I have more ideas but will 🤐 and keep it at that.

## ❤️ 

Shoutout to Reece, who has the sole search result providing a great template for this:

Screen Shot 2023-04-29 at 8 09 24 PM

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -382,6 +382,9 @@ func NewAppKeepers(
 		// governance
 		"/cosmos.gov.v1beta1.Query/Vote": &govtypes.QueryVoteResponse{},
 
+		// distribution
+		"/cosmos.distribution.v1beta1.Query/DelegationRewards": &distrtypes.QueryDelegationRewardsResponse{},
+
 		// staking
 		"/cosmos.staking.v1beta1.Query/Delegation":          &stakingtypes.QueryDelegationResponse{},
 		"/cosmos.staking.v1beta1.Query/Redelegations":       &stakingtypes.QueryRedelegationsResponse{},


Reecepbcups:
@mikedotexe I am all for getting more stargate queries in Thanks!

This actually has a standard query as well for a production ready solution now:
```rust
  use cosmwasm_std::{DelegationResponse, StakingQuery};

    let response: DelegationResponse = deps.querier.query(
        &StakingQuery::Delegation {
            delegator: env.contract.address.into_string(),
            validator: src_validator.clone(),
        }
        .into(),
    )?;
   ````
   
   We should probably just add a bunch / all the base cosmos messages Stargate queries soon

mikedotexe:
> This actually has a standard query as well for a production ready solution now:

Wow, thanks, I looked at protobufs for a while and missed this 🙏🏼 

Particularly the `pub accumulated_rewards: Vec<Coin>,` from `FullDelegation` is what I'm looking for

Reecepbcups (approved):
LGTM

faddat:
Thank you very much Mike

PR #666, opened 2023-05-06
use cosmossdk.io/errors

faddat:
This PR uses cosmossdk.io/errors, and will reduce the size of diffs 
when upgrading to sdk 47.


Reecepbcups:
So is the only reason for this PR to "reduce the size of diffs" for SDK v47? Just keep it in the 47 branch if so. No need to bring this minor change before then IMO

faddat:
that's the reason for it.  It doesn't add features but does improve maintainability a lot. 

merge or not as you wish ser (close if not desired)

Reecepbcups (approved):

Reecepbcups:
Closing as it will be done in the v47 PR (v15)

PR #676, opened 2023-05-15
Fix v47 IBC test

Reecepbcups:
Trying to narrow down where the issue lies for IBC not ack'ing

- Removes Ibchooks
- removes PFM

^ This fixes the channel-id not found error. But the packet is still not acked.

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -28,3 +28,5 @@ jobs:
           platforms: linux/amd64
           push: false
           build-args: arch=x86_64
+          cache-from: type=gha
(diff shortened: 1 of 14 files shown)


Reecepbcups:
:crossed_fingers:

Reecepbcups:
Issue looks to be with PFM then. Locally adding PacketForwardMiddleware back in causes the `Unknown desc = failed to execute message; message index: 0: channel-0:` issue again. 

So IBCHooks is good
Relayer is also good (though requires manual flushing during test)

Reecepbcups:
Seems IBCFee is not playing nicely with PFM as the ICS4Wrapper

PR #684, opened 2023-05-20
[v15] IBCFee test

Reecepbcups:
## Update In the current state of 29-fee I do not think it is testable

1) Allowing channel-0 at genesis for ibc-fee does not allow the channel to open properly when its time to init after genesis
2) This means it requires an upgrade handler, but this is not testing an upgrade - it is testing IBCFee e2e. So we would have to keep up with 2 binaries for every push, one without IBCFee upgrade handler, and one with. **Complexity^2**

---

Merged into the 47-upgrade branch. May point to main in the future after that merges depending on timelines

# Feature
Adds a ictest to ensure IBCFee (29-fee IBC-Go spec) is implemented properly

# Files of importance
- interchaintest/module_ibcfee_test.go

# Todo
- Fix STATE_INIT being stuck and not STATE_OPEN for transfer channel-1
- Add .github CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,6 +131,9 @@ ictest-tokenfactory: rm-testcache
 ictest-feeshare: rm-testcache
 	cd interchaintest && go test -race -v -run TestJunoFeeShare .
 
+ictest-ibcfee: rm-testcache
+	cd interchaintest && go test -race -v -run TestJunoIBCFee .
+
 # Executes a basic chain upgrade test via interchaintest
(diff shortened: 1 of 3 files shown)


Reecepbcups:
TODO: Use https://github.com/cosmos/relayer/blob/main/interchaintest/fee_middleware_test.go instead since this fee_version metadata is required. This is a WIP though so may not work still

really we should just not use IBCFee if this is the case unless the channel upgrade thing works permissionlessly from our side in the future

PR #691, opened 2023-05-23
[v14] Adds v14.2.0 upgrade handler

Reecepbcups:
Upgrade to add TokenFactory params for gas cost only. Required for contract developers & ICS-999 :smile: 

Tested manually as well since we do not have TF CI for <= v14

Will tag as **v14.2.0-alpha.1**


dimiandre (approved):

the-frey:
Question: this is state breaking, right? Hence the upgrade handler...

If so we should tag as v15. Upgrade handlers for major versions has generally been the policy thus far

the-frey:
Yeah, the having to shuffle major version names where every major version was an upgrade handler was part of the reason we abandoned code-names after lupercalia and unity. In general people are understanding about the fact that a version is subject to change, but also the level of specificity of vX.X.X is arguably not as important to a non technical person as simply "next release", "the release after next" etc

Always the possibility of confusion, however...

Reecepbcups:
The real fun part is ICtest :P i cri

PR #698, opened 2023-05-25
[v16] Migrate internal modules params

### LINKED ISSUE #694: remove x/params
Many modules we have still depend on x/params

We should remove all references to it and implement their own params store for each module. 

an example on how to do is here https://github.com/cosmos/cosmos-sdk/pull/12363

Reecepbcups:
Closes #694 

This PR will migrate & test our internal modules upgrading from x/params to doing it within the modules themselves. New modules (ex: drip) do not have to provide the legacy code for migration. We can remove after the upgrade.

Review starts at: eaf78911aae62e7d3ccab2144810e81f6255e394

# Migrations done for:
- TokenFactory (I think, need to check queries)
- x/Mint

# TODO:
- x/GlobalFee (add keeper for Tx, use new staking params for denom TODO. Ante handler is prob incorrect)
- log each migration to console t.Log to ensure all modules get transferred properly.
- fix test sim
- ensure real e2e upgrade actually passes & migrates all properly
- Migrate FeeShare & mint grpc query to use the keeper types

## Features
- Setup proto correctly for docker generation
- Add MsgUpdateParams to our modules
- Add v2 migration & go test
- Add upgrade handler for migration
- Fix proto issues with cosmos/gogoproto v1.4.10 (reverted to v1.4.8 per SDK)
- Runs ICTest every workflow now


Reecepbcups:
The ICtest failure here is something from Jacob's changes in that merged in version bump PR. Going to tackle this & migrating other modules tomorrow after v15 testnet.

Reecepbcups:
Going to do each in their own PR, this will be too confusing and large

PR #715, opened 2023-06-16
feat!: Adds all stargate queries for smart contracts

### LINKED ISSUE #714: On the necessity of `AcceptedStargateQueries` in Juno
## Background

I've been talking with some CosmWasm devs about what we want to build for upcoming hackathons. Some of the cutting-edge areas to explore involve having a smart contract be "self-aware" as well as aware of other contracts' details. I could see great value in having contracts have access to Stargate queries that return info about itself or other contracts. 

For instance, `ContractHistory` could be a valuable query in the future. 
https://github.com/CosmWasm/wasmd/blob/a4548ba17526027ced9cfd3fe783c4d66831f9e4/proto/cosmwasm/wasm/v1/query.proto#L97-L104

Currently, to allow this query message, we'd need to add it to `keepers.go` in Juno to the whitelist, like I did here: https://github.com/CosmosContracts/juno/pull/658/files

For context, I want to note that there's a history of caution around Stargate queries because of real non-determinism experienced by Juno from some jokester who halted the chain.

Here's a screenshot from [this Discord message](https://discord.com/channels/737637324434833438/737643344712171600/1040920787512725574) in the CosmWasm Discord.

Screen Shot 2023-06-15 at 10 16 13 AM

Seeing as it's been a comfortable period of time since we've seen a chain halt due to non-determinism from a CosmWasm query, and Confio has addressed the shortcoming by not allowing the specific error _message_ to be returned (which was deemed to be the root cause if I'm reading this correctly) perhaps it's time we reconsider whether we need a shortlist of allowed Stargate queries anymore. It's possible this was a smart, cautionary move, but this guard may become vestigial at some point in maturity. The question is whether that time is now.

Before ending this section, I want to mention that at the time of this writing, the Cosmos ecosystem at large has not figured out the implementation details for account abstraction, though we're all excited for it. I believe besides the `ContractHistory` example earlier, some of the most important Stargate queries we'll need are around authz and feegrant queries. This is where the sauce is gon' happen.

Do we want to individually add Stargate queries for authz? Or can we do a blanket allowance for certain modules?

## Suggested Design

I'd suggest we hold a conversation around if we need this `AcceptedStargateQueries` approach anymore.

## Acceptance Criteria

A good, solid conversation where people feel empowered and understand one another. **Purvis chapter 6 verse 66**.

## Timeliness

Let's build some dope shit at the hackathons coming up in a month. Would be really cool to join heads on a long-term **and** temporary solution so hackers can change the world in real-time in a few weeks.

Reecepbcups:
Closes #714 

TODO:
- add go test and/or e2e test with ic for EVERY one of these
- lint


dimiandre:
I'm hyped but also scared

We need to double check event returns to be ordered and determinism of each of these

Reecepbcups:
@dimiandre totally agree - I want to have a contract / e2e test which checks all of this with -race across 8+ vals on a network to ensure its safe before merge

Reecepbcups:
@mikedotexe Are you able to work on a contract for this? You have more rustc proto experience than I do - or maybe a contract which uses [Nois's Anybuf](https://github.com/noislabs/anybuf)?

---

Guidelines if so:

- We need a contract which has query endpoint as the module names.  (ex: '{"bank":{}}')  - for query gas reasons & narrowing down issues
- Each contract query would have multiple stargate queries inside it for said module (Bank would be https://github.com/CosmosContracts/juno/pull/715/files#diff-1ac782d2fab1d48ad7f7ef06cf2d58b37c99b6b4c98879a5ad86a985b73fad58R628-R639, use the env.contract_addr as the address to check)
- I do not think we have to care about the return values? Just that the query does not crash the chain

Then I can build a e2e test with 8 validators so we can ensure the chain does not halt for non-determinism.

mikedotexe:
> Are you able to work on a contract for this?

Hey! My honest answer is I will probably need to reduce my time after AwesomWasm and OsmoCon to catch up on energy I've lost during my CosmWasm marathon. Maybe we can chat about this in late August?

mikedotexe:
> I'm hyped but also scared

I feel the same way. If it's preferable, perhaps we can consider adding a handful of Stargate queries that might unlock a lot of potential. I know Yieldmos is keen on having Authz and FeeGrant queries going. Those might be worthwhile to get in before  a larger, blanket acceptance of all Stargate queries?

Reecepbcups:
May be able to use Anybuf to make this easier to generate

https://github.com/noislabs/nois-contracts/blob/v0.13.6/contracts/nois-payment/src/contract.rs#L112-L118

app/keepers/keepers.go
```diff
@@ -468,31 +471,42 @@ func NewAppKeepers(
 	wasmOpts = append(wasmOpts, tfOpts...)
 
 	// Stargate Queries
+	// (Define custom modules here, then the next function takes care of the standard SDK types)
 	accepted := wasmkeeper.AcceptedStargateQueries{
+		// wasm
+		"/cosmwasm.wasm.v1/Query/ContractInfo":       &wasmtypes.QueryContractInfoResponse{},
```
Reecepbcups, line 477: These should be .v1.Query/ oops

Reecepbcups:
Going to have to postpone for a future upgrade, don't have time atm to write the test to ensure these all work.

app/keepers/keepers.go
```diff
@@ -614,6 +626,118 @@ func NewAppKeepers(
...
+
+		// upgrade
+		"/cosmos.upgrade.v1beta1.Query/CurrentPlan":                 &upgradetypes.QueryCurrentPlanResponse{},
+		"/cosmos.upgrade.v1beta1.Query/AppliedPlan":                 &upgradetypes.QueryAppliedPlanResponse{},
+		"/cosmos.upgrade.v1beta1.Query/QueryModuleVersionsResponse": &upgradetypes.QueryModuleVersionsResponse{},
+		"/cosmos.upgrade.v1beta1.Query/QueryAuthorityResponse":      &upgradetypes.QueryAuthorityResponse{},
+	}
+
+	updated := asq
+	for k, v := range std {
```
Reecepbcups, line 735: This is fine since we are iterating all keys & not exiting early (else we would need an ordered map with []string{}'s deterministic iter

elix1er:
M8 can we utilize and write automatization to push release protobufs to buf.build

Brain blast.. How about we automate our protobuf push to buf.build with GitHub Actions?

Here’s the plan:

1. Trigger it on a branch update or a new release - we'll nail down which one.
2. Run `protoc` in the workflow to handle our protobuf compilation.
3. Using our Go files (or some IBC-compatible state machine magic), beam those bufs straight to buf.build.

Whatcha think? Straightforward enough, right?

Let's roll cuz i need this in my data sink pipe magic workflow
If not exist let me know im gonna make proof of concept raw

Would benefit not only me. Some chains up there already, many not and incomplete AF.

elix1er:
#!/bin/bash

# Install buf.build
# (This might be platform-specific, ensure you have the right installation instructions.)
# For example purposes, I'll use their standard install command from the website.
curl -sSL https://install.buf.build/install.sh | sh

# Set up a config for a new repository
buf mod init buf.build/juno/juno-core

# Login to buf registry
buf registry login buf.build --username 'elix1er'

# Prompt the user for manual token input
echo "Please create a token and paste it into the command line to complete the login."

# Push the module
buf push

dimiandre:
> 
> * I do not think we have to care about the return values? Just that the query does not crash the chain
> 

@Reecepbcups return values must be deterministic and in same order if array, same for return events

faddat:
@dimiandre is right, this is something we should be hyped and scared about, it needs tests.

PR #720, opened 2023-06-18
faddat/gci2

faddat:
This PR introduces gci, a linter that deterministically orders the imports in Juno.

- gci config for juno
- use gci linter on juno for deterministic import ordering


dimiandre (approved):
I don't see why not!

Thanks jacob

Reecepbcups (approved):
LGTM

PR #729, opened 2023-06-23
Faster ICTest block times

reecepbcups:
- From 2s to 750ms now. Helps to speed up the IBC testing
- Ignores some more dirs for faster docker build times


Reecepbcups:
tbh not worth the risk of faster blocks.

PR #730, opened 2023-06-23
[v16] TokenFactory Validation Migration

### LINKED ISSUE #719: Bug: Tokenfactory exports not a valid genesis
Tokenfactory has a bug where exports are not a proper genesis due to [x/bank metadata validation](https://github.com/cosmos/cosmos-sdk/blob/v0.47.2/x/bank/types/metadata.go#L11). Only affects hard forks.

We should have a export test in the future with ictest to ensure hard forks are possible on the genesis & is valid.

Thanks to @philipsu522 for the ping, will credit in the commit & release notes.

---

Fix PR: https://github.com/sei-protocol/sei-chain/pull/861

Migration Logic: https://github.com/sei-protocol/sei-chain/blob/master/x/tokenfactory/keeper/migrations.go#L44

Reecepbcups:
Closes #719
Per https://github.com/sei-protocol/sei-chain/pull/861

When new tokens are created, some of the bank metadata is not set by default. This causes x/bank's genesis validation to fail on a hard fork of the mainnet genesis.

Task:
- Migrate all TF denoms without Display to have the required metadata
- Test to ensure different token-factory states properly convert (e2e upgrade)
    - denom, no supply
    - denom with supply
    - denom with supply with set metadata (ex: `junod tx tokenfactory modify-metadata` already)


dimiandre (approved):

0xstepit (approved):
LGTM

PR #735, opened 2023-06-28
Move Wolf's Vesting account -> Core-1 subDAO

Reecepbcups:
Runs the following when chain-id == "juno-1":

- Unlock all tokens (end time of epoch time 1)
- Force finish any redelegations
- Force Unbond all current delegations
- Get balance
- Send to Core1 subDAO

TODO:
- Need to add standard logic checks to ensure this works properly. Can re-use for all Core-1 -> vesting contracts.
- Just do it for all Core-1 members from the other PR? (And init contracts on their behalf)

### DIFF
--- a/app/upgrades/v16/upgrades.go
+++ b/app/upgrades/v16/upgrades.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
+
 	// SDK v47 modules
 	// minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
@@ -30,6 +31,7 @@ import (
 
 	"github.com/CosmosContracts/juno/v16/app/keepers"
 	"github.com/CosmosContracts/juno/v16/app/upgrades"
+
 	// Juno modules
 	feesharetypes "github.com/CosmosContracts/juno/v16/x/feeshare/types"
 	globalfeetypes "github.com/CosmosContracts/juno/v16/x/globalfee/types"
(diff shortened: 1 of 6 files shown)


Reecepbcups:
Upgrade only fails now due to said account not being a Vesting account in the local e2e.

Need to add standard logic checks to ensure this works properly. Can re-use for all Core-1 -> vesting contracts.

dimiandre:
It shouldn't move all the funds but only the ones that are still waiting for vesting, also better to not delete the account

the logic should do:

1- count how many tokens are still in vesting period
2- unstake it
3- move to core 1 treasury

Reecepbcups:
Thanks @dimiandre, I was overthinking it for sure. Logic is much cleaner now & follows this better. Tested locally and is a success.

Core-1 migration can use the same logic, just need to instantiate contracts on behalf of the SubDAO.

Using mock account of
```
'{"@type":"/cosmos.vesting.v1beta1.PeriodicVestingAccount","base_vesting_account":{"base_account":{"account_number":"0","address":"juno1xz599egrd3dhq5vx63mkwja38q5q3th8h3ukjj","pub_key":null,"sequence":"0"},"delegated_free":[{"amount":"2000","denom":"ujuno"}],"delegated_vesting":[{"amount":"0","denom":"ujuno"}],"end_time":"1701193279","original_vesting":[{"amount":"2501","denom":"ujuno"}]},"start_time":"1688233279","vesting_periods":[{"amount":[{"amount":"1250","denom":"ujuno"}],"length":"0"},{"amount":[{"amount":"1251","denom":"ujuno"}],"length":"12960000"}]}'    
```

```
== Vesting Account Address: juno1xz599egrd3dhq5vx63mkwja38q5q3th8h3ukjj ==
Locked / waiting to vest Coins: 1251ujuno
Vesting Account set to BaseAccount, not more vesting periods.
Minting Unvested Coins back to Core-1: 1251ujuno
1:17PM INF minted coins from module account amount=1251ujuno from=mint module=x/bank
Updated Core1 SubDAO Balance: 1251ujuno
```

Reecepbcups:
Hmm, May make more sense to just review PR 741 instead of doing this in 2 goes since its the same logic, just a iterator + bool check

PR #741, opened 2023-07-03
migrate: Core-1 vesting accounts to the Charter

Reecepbcups:
## Summary
Migrates all Core-1 to new vesting contracts. If the wallet is wolf, send funds to the Core-1 subdao (No more vesting for him)

Uses - https://pastebin.com/raw/89MJms1c - to test manually

SDK v47 now lets us create accounts at runtime

Should also use Osmosis as ref: https://github.com/osmosis-labs/cosmos-sdk/blob/osmo/v0.47.5/x/auth/vesting/types/vesting_account.go#L813-L859


app/upgrades/vesting.go
```diff
@@ -0,0 +1,273 @@
...
+	// Clears the account so all all future vesting periods are removed.
+	clearVestingAccount(ctx, vacc, keepers, unvestedCoins)
+
+	// Complete any re-deleations to become standard delegations.
+	if err := completeAllRedelegations(ctx, keepers, accAddr, now); err != nil {
+		return err
+	}
+
+	// Instant unbond all delegations.
+	_, err := unbondAllAndFinish(ctx, now, keepers, accAddr)
```
dimiandre, line 73: Why do we need to undelegate everything?

Reecepbcups: tldr; hassle and minimizing risk

1) I clear the vesting account so that all vesting periods are unlocked. If a Core-1 member has more than 51% of vested tokens delegated, it's a nightmare to do the math. 
This reduces risk of incorrect math and just ensures all tokens are undone before trying to bankSend -> the Core-1 DAO -> vesting contract init.

2) All these vested tokens are moving to the new vesting contracts so we have to unvest them. I can't move delegations to the new contracts until they are created, which can only be done after I unvest all funds and return them to the base account.

Then Core-1 members can re-stake after the upgrade in the new CW vesting UI
_(I will get the withdraw/recipient addresses from you guys in the coming days)_

**NOTE**: Any rewards / vested funds which a Core-1 member has in their SDK vesting account will STAY in said account since they are the controller of it / it has been unlocked.

Reecepbcups:
Bumping to v17+ due to complexity. Would rather it be safe and get more testing since we are moving like millions in funds

Reecepbcups:
TODO: move into the v17 upgrade handler

Reecepbcups:
Needs a signaling proposal via main gov, not just Core-1 per validator comments to me.

Reecepbcups:
Putting this snippet here from the other PR to remove the multisig, as it may be useful

```
func mintUnvestedToCharter(ctx sdk.Context, k *keepers.AppKeepers, CouncilAddr sdk.AccAddress, vestingAcc *vestingtypes.PeriodicVestingAccount) { // nolint:gocritic
	unvested := SumPeriodVestingAccountsUnvestedTokensAmount(ctx, vestingAcc)
	fmt.Printf("Core1Addr Unvested to mint to the charter: %s\n", unvested)

	coins := sdk.NewCoins(sdk.NewCoin("ujuno", unvested))

	if err := k.BankKeeper.MintCoins(ctx, "mint", coins); err != nil {
		panic(err)
	}

	if err := k.BankKeeper.SendCoinsFromModuleToAccount(ctx, "mint", CouncilAddr, coins); err != nil {
		panic(err)
	}
}
```

app/upgrades/v19/upgrades.go
```diff
@@ -14,6 +14,48 @@ import (
...
+	{
+		Owner:   JackKey,
+		Address: "juno130mdu9a0etmeuw52qfxk73pn0ga6gawk4k539x",
+	},
+	{
+		Owner:   "jake",
+		Address: "juno18qw9ydpewh405w4lvmuhlg9gtaep79vy2gmtr2",
+	},
+	{
+		Owner:   "multisig",
```
Reecepbcups, line 50: no longer needed for v20, done in v19

joelsmith-2019: Good catch, removed in v20 handler

app/upgrades/v19/upgrades.go
```diff
@@ -50,6 +68,31 @@
...
+		if err := upgrades.MoveVestingCoinFromVestingAccount(ctx,
+			keepers,
+			bondDenom,
+			name,
+			sdk.MustAccAddressFromBech32(vestingAccount),
+			sdk.MustAccAddressFromBech32(Core1SubDAOAddress),
+		); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
joelsmith-2019: Resolved with the use of an array

dimiandre:
With Ops Department we decided to put this on hold since not all core-1 members agreed to cancel their vesting

faddat:
probably should be closed, unless the migration isn't actually done

Reecepbcups:
not done, only the multisig was transfered. doubt it will get through since the charter is against upgrades

faddat:
thanks for clarifying @Reecepbcups

PR #743, opened 2023-07-05
fix(test): ICTest relayer fix + remove redundant PFM

Reecepbcups:


Reecepbcups:
Probably need to wait for this now ugh
https://www.mintscan.io/juno/proposals/307

If this passes I need to upgrade the handler to use the new values (since x/params legacy subspace had issues with GlobalFee)

faddat:
Strange place to make a change to security.md

though I do agree with the change

Reecepbcups:
It was found during cleanup of the upgrade, so it was... cleaned up.

Visibility is in the commit & the text of the description. also assumed from that public leaving statement you gave

PR #692, opened 2023-05-23
[v16] SDK v47 Cleanup branch

Reecepbcups:
Clean up minor nits from the review session & some CI things.

## Features
- Comments out test-simulation.yml workflow for now (Will add back after we migrate modules over)
- Removed Unused FeeShare functions
- Improves the test_node.sh script
- Upgrades from v15 namespace to v16
- Adds 2 new stargate queries
- Adds the NFT module
- IBC Bump (Huckle)
- v0.40.0 wasmd

# Other PRs which are a WIP
- migrate module params
- upgrade handler (Double check IBC & re-read guides from SDK migrations & wasmd)

PR #746, opened 2023-07-09
remove myself from security.md in a visible way

faddat:
This PR removes me from Juno's security contacts in a visible way, instead of being bundled into another changeset as in #692


Reecepbcups:
It is visible in the other PR pending a review & merge

PR #758, opened 2023-07-18
[v17] feat!: FeeShare allow DAODAO to register contracts

### LINKED ISSUE #593: FeeShare v2
1) Emit event within the ante handler for the amount of fees split for an execute?

~~2) A few have asked, Allow for a contract to specify minimum fees on top of the original Tx
The issue here is the UX from keplr / wallet side. We could add a query to get this from the FeeShare keeper, then add that fee to the keplr Tx. - Is it even worth it? As a contract dev, just add a required `--amount` sent to the contract. Then require this for all contract executes~~
yea no this "Fee premiums" idea is dumb, CosmWasm already has it built in

3. Allow code id user to register all sub contracts (swift wants) which are instantiated. This would override any currently registered types of said code id
    - On initial register allow for the code_id uploader to specify: can be overridden by instantiators OR no override
    
4. Keep up with all fees generated per contract (in upgrade handler, set all current to a value of 0)

Reecepbcups:
Closes #593

- Allows for a DAODAO dao to register a contract it is the admin of.  Before the contracts could only be registered to themselves.
- On any execute, emits a JSON event for the feeshare payouts. Useful for indexing in the future

```
- attributes:
  - index: true
    key: payouts
    value: '[{"withdraw_address":"juno1v75wlkccpv7le3560zw32v2zjes5n0e7csr4qh","fees_paid":[{"denom":"ujuno","amount":"5000"}]}]'
  type: payout_feeshare
 ```

Cleanups up some logic + adds test.


dimiandre (approved):
How disk space will be affected? 

I can imagine majority of the contracts calls transactions will be with feeshare payout event attached

Reecepbcups:
@dimiandre for the event emit? basically nothing relative to the size/compute of CW executes

I will benchmark the gas cost difference tomorrow, but I expect it to be <500 gas if even

Reecepbcups:
@dimiandre 
    # Execeuted X times on a fresh chain, same specs, with feeshare events
    # gas_used: "138041"
    # gas_used: "138048" * 10
    # Executed X times on a contract WITHOUT Feeshare event (feeshare still registered)
    # gas_used: "138_041"
    # gas_used: "138_048" * 10
    
   So no gas changes. The string is ~200bytes, so cost <5MB every 25mil contract calls WITH feeshare enabled + any CosmosSDK indexing overhead, but small in comparison

PR #761, opened 2023-07-19
fix: add genesis subcommand and gas prices for local docker

shyambin:
- add genesis subcommand in docker container shell script while staking an amount and collecting gentxs

### DIFF
--- a/docker/run_junod.sh
+++ b/docker/run_junod.sh
@@ -6,4 +6,4 @@ if test -n "$1"; then
 fi
 
 mkdir -p /root/log
-junod start --rpc.laddr tcp://0.0.0.0:26657 --trace
+junod start --rpc.laddr tcp://0.0.0.0:26657 --minimum-gas-prices 0.0001ucosm --trace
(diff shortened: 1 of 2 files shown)


docker/run_junod.sh
```diff
@@ -6,4 +6,4 @@ if test -n "$1"; then
 fi
 
 mkdir -p /root/log
-junod start --rpc.laddr tcp://0.0.0.0:26657 --trace
\ No newline at end of file
+junod start --rpc.laddr tcp://0.0.0.0:26657 --minimum-gas-prices 0.0001ucosm --trace
```
Reecepbcups, line 9:
```suggestion
junod start --rpc.laddr tcp://0.0.0.0:26657 --minimum-gas-prices 0.0001ujunox--trace
```

Reecepbcups:
@shyambin Thanks for this! We should really have had a test to ensure this works on new releases. I am just going to change it from ucosm gas price to ujunox per the docker-compose file

Note: in the future this will be removed in favor of using https://github.com/Reecepbcups/local-interchain which is a much better experience (easy IBC, run multiple juno chains on Windows, MacOS, Linux, etc)

PR #774, opened 2023-08-03
fix: Fix pointer to store upgrades inside loop

outofforest:
Due to how `for` loop works, if many upgrade plans are specified, the store upgrades are taken from the last one on the list, not the one matched by the upgrade name. To fix this, the copy of the object is made, before taking the pointer.


Reecepbcups:
nice catch! I'll test around with this today & get it in after

Reecepbcups (approved):
Works as expected, thank you very much for this!

PR #777, opened 2023-08-08
fix: typo run_junod.sh

whalelephant:


the-frey:
@Reecepbcups is E2E failing here because it's an outside contributor?

Reecepbcups:
@the-frey Yes, 403 forbidden on the docker push even though I gave the CI write access to the repo

the-frey:
Vaguely remember this having been an issue elsewhere with some CI workflows. I *guess* on merge it should be a co-commit from you/me so it will pass CI 🤷

the-frey (approved):

PR #789, opened 2023-08-14
feat!: Faster block times (3s)

### LINKED ISSUE #286: Lower block times
### Highly informal presentation on Juno block times

**Problem:**

- we want fast blocks
- validators miss lots of blocks because we want fast blocks
- validators feel odd about missing lots of blocks because of the visualization of red streaks across the web on block explorers like ping.pub

**Solution**

- don't change the commit timeout value, it should remain at 5s
- use skip_timeout_commit=true — this will cause Juno to make blocks as fast as possible, but vals will wait for other signers.  The maximum block time will remain the same as it is now. 
- If a validator is down, block times will get longer.  This will increase the pressure to do a great job of validating, because downtime will slow osmosis down.

**Context**:
https://twitter.com/blockpane/status/1572823202753089536

**Desired result**

This should allow osmosis to produce blocks very raipidly without causing osmosis validators to miss significantly more blocks. 

When a validator is down, osmosis will produce blocks at its current speed.  Theoretically speaking, this reactivity will increase the pressure on validators to stay up.

Reecepbcups:
Closes #286

3s block times, tested locally as well and confirmed working as expected
upgrade article will include CosmosCache + setup + support to better handle request scalability 

## Summary
- Doubles slashing window
- Doubles block per year minted (since blocks are 2 times faster
- Forces consensus timeout commit to be 2 seconds (+1 second tendermint p2p = 3 seconds)
- test params update properly via go test


elix1er:
Lit

Reecepbcups:
Bug in CometBFT seems to now actually allow overwriting Tendermint Config's :( so the upgrade is successful, but having issues overwriting the actual timeout now

dimiandre (approved):

PR #796, opened 2023-08-15
ictest: cache deps on setup, cache Docker Image layers

### LINKED ISSUE #790: e2e: download and cache go deps
For interchaintest-e2e.yaml github action

This way each sub task in the matrix does not have to re install it. Should lower test times by ~3 minutes each (esp. helpful for re-runs on github CI failure)

Only `interchaintest/go.sum` should need the cache for it, since the build is only done once for the root go.sum

Reecepbcups:
Closes #790

Get other references from repos such as the tokenfactory-contracts and other e2e external repos


Reecepbcups:
does not work as expected. Will try again in the future, not a priority

PR #797, opened 2023-08-17
feat!: x/clock module

Reecepbcups:
## Summary
The new x/clock module auto 'ticks' smart contracts which are whitelisted in

## Files of Importance
- x/clock/abci.go

## TODO
- [x] Logic & Design
- [x] test
- [x] e2e with actual param prop
- [x] Spec sheets


x/clock/abci.go
```diff
@@ -0,0 +1,42 @@
...
+
+// EndBlocker executes on contracts at the end of the block.
+func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
+	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyEndBlocker)
+
+	message := []byte(types.EndBlockSudoMessage)
+
+	p := k.GetParams(ctx)
+
+	errorExecs := make([]string, len(p.ContractAddresses))
```
Reecepbcups, line 22: Contracts should really never error unless they either 
a) Do not exist yet (useful for testing
b) Migrate and remove / alter the original SudoMsg they had before original upload.

x/clock/abci.go
```diff
@@ -0,0 +1,42 @@
...
+			continue
+		}
+
+		_, err = k.GetContractKeeper().Sudo(ctx, contract, message)
+		if err != nil {
+			errorExecs[idx] = addr
+			continue
+		}
+
+	}
```
Reecepbcups, line 37: Do we want to remove any used gas here from the GasMeter? (Get GasMeter before we iterate, then after each sudo call we get the difference (new-old) and subtract that from new)

app/keepers/keepers.go
```diff
@@ -563,6 +570,13 @@ func NewAppKeepers(
 		govModAddress,
 	)
 
+	appKeepers.ClockKeeper = clockkeeper.NewKeeper(
+		appKeepers.keys[clocktypes.StoreKey],
+		appCodec,
+		*appKeepers.ContractKeeper,
+		govModAddress,
```
Reecepbcups, line 577: Do we want a SubDAO to control contracts instead of Gov? (Quicker, techy people vs just token holders)

x/clock/abci.go
```diff
@@ -28,12 +28,12 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
 			continue
 		}
 
-		_, err = k.GetContractKeeper().Sudo(ctx, contract, message)
+		childCtx := ctx.WithGasMeter(sdk.NewGasMeter(types.EndBlockContractGasLimit))
```
Reecepbcups, line 31: I like this more as a safety check since it is in the endblock anyways. (1bn gas was used by terra pre collapse, and in my testing a VPS node can support it in the endblocker no problem)

app/keepers/keepers.go
JakeHartnell: Yes! This would be a great improvement.

x/clock/abci.go
Reecepbcups: May want to move this to a gov param

Reecepbcups: Added to gov params 83d51d2dba40ca73bfd82c56c4a77ecb6075bf8f

Reecepbcups:
renamed to clock_end_block so as to not cause issues with twasm in the future when we add that

app/keepers/keepers.go
dimiandre: You can implement an allowlist similar to x/drip

dimiandre (approved):

app/keepers/keepers.go
Reecepbcups: I agree

I think for now we should just use standard Gov account then figure out the proper SubDAO (mainnet, and testnet?) in a v2 of the module

PR #801, opened 2023-08-21
[v18] Juno x/burn module

### LINKED ISSUE #647: x/burn 🔥
# `x/burn` 🔥

## Background

The Juno community wants a canonical burn process. Currently burning causes `x/mint` to redistribute burned amounts. 

> If we burn it the mint module would remint the amount and giving it to the stakers yes ? 

> thats how it currently works, the 2nd year would basically be longer and wouldnt end in october

## Suggested Design

`x/burn` module. The module has an address where tokens and other assets can be canonically burned.

## Acceptance Criteria

- [ ] PR with a sound and functional `x/burn` module.
- [ ] Signalling proposal accepted by governance that indicates the desire to declare a canonical burn address.

Reecepbcups:
Closes #647

## Files of importance
- x/burn/* (more of a wasm plugin override than a module)
- app/keepers/keepers.go -> wasmkeeper.WithMessageHandler

## Usage
There are 2 ways to burn tokens
1. Send directly to the module account address `juno1mj7t69y4r2adl3cnuq8y9uundkzawvx6avu7nj`
2. Bank::Burn in CosmWasm

## TODO
- [x] e2e test https://github.com/dimiandre/cw-testburn
- [x] once above is done, remoev old MessageHandlerDecorator in keepers.go


dimiandre:
This way the query "juno q bank total" will be wrong and It's used by many third party services. 

I'm not sure if this is the proper way to do it. 

Probably better to keep the standard burn functions and modify our mint module so that it's also reduced in minter.TargetSupply

Reecepbcups:
TODO: Voting yes on the Unity contract continues to fail for not passing vote. Very odd its effecting SudoContracts

dimiandre:
@Reecepbcups I pushed a new commit. 

The burning was not working before because coins were sent to x/junoburn module but burn was happening on x/wasm

My commit:

1. Removes burn permissions to x/wasm. This is to prevent other part of wasm handing burns in a not appropriate way. For example: https://github.com/CosmWasm/wasmd/blob/6a2bffd479acd212d76ac92681b1b6ec099e649d/x/wasm/keeper/keeper.go#L1188
2. Give burn permission to junoburn module/plugin
3. Fix module names, tokens are now transfered & burned from junoburn and never reaching wasm account
4. Remove the total supply adjustment in mint module, it's not needed because we are modifying the target supply

dimiandre (approved):
lgtm! 

great job with contract refactor and tests!

PR #805, opened 2023-08-22
chore!: rename to TokenFactoryMsg (match osmosis)

Reecepbcups:
Ref: https://github.com/CosmosContracts/token-bindings/pull/5

## Description
Makes it easier to deploy the same TF contract on Osmosis & Juno with matching bindings _(will break all TF contracts. Requires a cargo version bump)_

Tested locally with [local-interchain & Jake's script](https://gist.github.com/Reecepbcups/e47eea449ecddbcc19b62ec76afc44f4) as well


JakeHartnell (approved):
THANK YOU! Nice to make TF actually interoperable.

PR #809, opened 2023-08-24
Prettier GlobalFee fees errors

Reecepbcups:
I have seen a lot of developers confused by the CLI fee message (all of cosmos, not just Juno). This makes it prettier so it's easier to understand

---

not enough
image

not accepted
image

none input at all but one is required
image

PR #810, opened 2023-08-24
Fix: Prettier GlobalFee commit error

Reecepbcups:
Um what...

I pushed 27301e604f35d769b77892aea2cdfbd942958752 up. It is in the branch [reece/pretty-globalfee](https://github.com/CosmosContracts/juno/commits/reece/pretty-globalfee), but the PR #809 did not actually capture it. 

Really confused how that could have happened. Juno broke GH lol


JakeHartnell (approved):
LGTM

PR #816, opened 2023-09-01
reduce tokens in targetsupply when burning

dimiandre:
this pr is targeting reece branch

- This way the bank.totalSupply will always be right
- By reducing target supply, we reduce also the maxSupply and we shorten the halving date

maxSupply is actually reduced way more than the burned coins, because maxsupply is calculated dynamically at each 1 year epoch based on inflation schedule

this table estimates how much the max supply will be impacted for a burn in each phase schedule

|  Year | Burn Amount   | Effective Max Supply reduction   |
|---|---|---|
|  2 | 1 JUNO | 2.04 JUNO  |   
| 3  | 1 JUNO |  1.70 JUNO |   
| 4  | 1 JUNO |  1.54 JUNO |
| 5 | 1 JUNO | 1.41 JUNO |
| 6 | 1 JUNO | 1.31 JUNO |
| 7 | 1 JUNO | 1.22 JUNO | 
| 8 | 1 JUNO | 1.15 JUNO |
| 9 | 1 JUNO | 1.10 JUNO |
| 10 | 1 JUNO | 1.06 JUNO | 
| 11 | 1 JUNO | 1.03 JUNO |
| 12 | 1 JUNO | 1.00 JUNO |

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -525,7 +525,7 @@ func NewAppKeepers(
 	// 	return wasmkeeper.NewBurnCoinMessageHandler(customBurner)
 	// })
 
-	junoBurnerPlugin := junoburn.NewBurnerPlugin(appKeepers.BankKeeper)
+	junoBurnerPlugin := junoburn.NewBurnerPlugin(appKeepers.BankKeeper, appKeepers.MintKeeper)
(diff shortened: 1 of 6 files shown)


Reecepbcups:
Thanks! This looks good to me. Something with modifying Mint/Burn in both branches seems to break the Unity prop. So ill fix that in the other branch & add a Burn contract tester

PR #817, opened 2023-09-01
[v18] add query for x/mint target supply

dimiandre:
should be not consensus breaking


x/mint/client/cli/query.go
```diff
@@ -116,3 +117,31 @@ func GetCmdQueryAnnualProvisions() *cobra.Command {
...
+			}
+			queryClient := types.NewQueryClient(clientCtx)
+
+			params := &types.QueryTargetSupplyRequest{}
+			res, err := queryClient.TargetSupply(cmd.Context(), params)
+			if err != nil {
+				return err
+			}
+
+			return clientCtx.PrintString(fmt.Sprintf("%s\n", res.TargetSupply))
```
Reecepbcups, line 140: I think we should put it as the denom here too in the query

ex: `140015400000` is not as explicit as `{"denom":"ujuno","amount":"140015400000"}`

Reecepbcups:
deps #818

Reecepbcups (approved):

PR #821, opened 2023-09-04
DNM: [uni-a] Standalone Upgrade Patch for Testnet

Reecepbcups:
Upgradename: `uni-a`

Will be a Uni only tag off of a `releases/testnet-only` branch (based off of v17)

Mints Reece 100m tokens for testing and distribution


app/upgrades/uni-a/upgrades.go
```diff
@@ -0,0 +1,64 @@
...
+
+		// Run migrations
+		versionMap, err := mm.RunMigrations(ctx, cfg, vm)
+		if err != nil {
+			return nil, err
+		}
+
+		// Turn off inflation
+		minter := keepers.MintKeeper.GetMinter(ctx)
+		minter.Inflation = sdk.NewDecWithPrec(0, 2)
```
Reecepbcups, line 42: @dimiandre this would work since right?
- inflation is 0, so no new tokens are created,
- thus no new phases to alter inflation

dimiandre: Yes this should work, mint beginblocker is skipped almost immediately if inflation is equal to 0

dimiandre (commented):
1M tokens doesn't seems enough at a glance to stabilize the testnet, or is it? 

Also, maybe we should put this on a separate branch instead of main? and tag a uni only version. 

It's not a code we'll keep and I'd prefer to not mess the main codebase with weird token adjustments

Reecepbcups:
> 1M tokens doesn't seems enough at a glance to stabilize the testnet, or is it?
> 
> Also, maybe we should put this on a separate branch instead of main? and tag a uni only version.
> 
> It's not a code we'll keep and I'd prefer to not mess the main codebase with weird token adjustments

Oops! Was meant to be 100m thanks for the catch. should have been 1e9 not 1e6

Sounds good on the branch part. We could just tag it off here and not merge to main.

Reecepbcups:
GH workers are broken for now, nice

PR #827, opened 2023-09-06
bump cosmos iavl for v16.x

Reecepbcups:
backport of #826 (origin offer matt) -> v16 release branch

running on my testnet valiadtor & mainnet rpc nodes now

- golanglint failure is due to an updated lint check in the latest version, this is fixed in v17 but did not apply to v16 at the time. Leaving unchanged


dimiandre:
tested on my node, not consensus breaking ✅

PR #833, opened 2023-09-10
[v18] FeeShare: Allow authz exec to count towards feeshare payouts

Reecepbcups:
Thanks to Aleem from Vitwit for pointing this out :)

## summary
Allows feeshare authz executes to pay the wasm contract its % of feeshare rewards.
(Reuses logic from past decorator in v14.0.0)


JakeHartnell (approved):

x/feeshare/ante/ante.go
```diff
@@ -62,29 +63,63 @@ type FeeSharePayoutEventOutput struct {
 	FeesPaid        sdk.Coins      `json:"fees_paid"`
 }
 
+func addNewFeeSharePayoutsForMsg(ctx sdk.Context, fsk FeeShareKeeper, toPay *[]sdk.AccAddress, m sdk.Msg) error {
+	if msg, ok := m.(*wasmtypes.MsgExecuteContract); ok {
```
aleem1314, line 67: If the message contains another `authz.MsgExec` this logic doesn't work. You should add additional case for `authz.MsgExec`, and if it contains another `authz.MsgExec` message, call the function recursively.

x/feeshare/ante/ante.go
```diff
@@ -62,29 +63,63 @@ type FeeSharePayoutEventOutput struct {
...
-	for _, msg := range msgs {
-		if _, ok := msg.(*wasmtypes.MsgExecuteContract); ok {
-			contractAddr, err := sdk.AccAddressFromBech32(msg.(*wasmtypes.MsgExecuteContract).Contract)
+
+	// validAuthz checks if the msg is an authz exec msg and if so, call the validExecuteMsg for each
+	// inner msg. If it is a CosmWasm execute message, that logic runs for nested functions.
+	validAuthz := func(execMsg *authz.MsgExec) error {
+		for _, v := range execMsg.Msgs {
+			var innerMsg sdk.Msg
+			if err := json.Unmarshal(v.Value, &innerMsg); err != nil {
```
aleem1314, line 100: I think this is always going to fail. You should use SDK codec to correctly unmarshal any type to an interface.

joelsmith-2019: To add on to what aleem is saying, It turns out authz.MsgExec has a built in function called "GetMessages()" which returns an arrray of sdk.Msg. Way easier than looping through and trying to unmarshal.

PR #842, opened 2023-09-21
[v18] latest pkg versions

Reecepbcups:
## version bumps & minor tweaks 

- wasmd v0.42
- wasmvm 1.4.1


Reecepbcups:
no, bad idea for now

PR #849, opened 2023-10-04
Bypass signature verification for testing

Reecepbcups:
Good use case would be for example, interchaintest.

Take mainnet state in (simulate a whale) and perform actions on their behalf locally against what would actually happen & with the accounts which would be interacting.

### DIFF
--- a/app/ante.go
+++ b/app/ante.go
@@ -1,6 +1,8 @@
 package app
 
 import (
+	"os"
+
 	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
 	wasmtypes "github.com/CosmWasm/wasmd/x/wasm/types"
 	builderante "github.com/skip-mev/pob/x/builder/ante"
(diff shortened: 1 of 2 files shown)


dimiandre (changes requested):
this doesn't seem very safe to me, it opens more weak spots on the node for a coordinated attack. Maybe better to keep it in a separate tag/branch

Reecepbcups:
@dimiandre it would still require 51% of the VP to be compromised to run this attack. I see it as the same security risk as using Cosmovisor would be, no?

dimiandre:
> @dimiandre it would still require 51% of the VP to be compromised to run this attack. I see it as the same security risk as using Cosmovisor would be, no?

Cosmovisor is not installed by default, this is an option available to anyone and a simple os env var can activate it. Os vars can be very easily manipulated by any external software. 

Also i think with just 33% of the vp this can cause a chain halt

PR #850, opened 2023-10-04
revert wasmvm to 1.3 (from 1.4)

Reecepbcups:
per: https://x.com/CosmWasm/status/1709507168448229497?s=20

We should wait on 1.4.1 / wasmd v43. I am having some SDK issues using it, not sure why. Just do not feel comfortable moving. Will do for v19/20


Reecepbcups:
Ill bump us to 42 and 1.4.1. Still scares me so I need to add conformance test in the upgrade handler

JakeHartnell (approved):
Good call.

PR #871, opened 2023-10-23
[v18] wasm 1.4/wasmd 43, ibc+pfm latest, cw upgrade validation

Reecepbcups:
- wasmd v0.43
- wasvm 1.4.0
- ibc-go v7.3.1
- pfm latest
- Upgrade test: validate CW works & does not panic after upgrade (cc: Chihuahua chain halt due to wasmer issue)

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -195,7 +195,6 @@ func NewAppKeepers(
 	bApp *baseapp.BaseApp,
 	cdc *codec.LegacyAmino,
 	maccPerms map[string][]string,
-	enabledProposals []wasmtypes.ProposalType,
 	appOpts servertypes.AppOptions,
 	wasmOpts []wasmkeeper.Option,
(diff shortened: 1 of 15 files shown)


app/keepers/keepers.go
```diff
@@ -606,11 +605,6 @@ func NewAppKeepers(
 		),
 	)
 
-	// register wasm gov proposal types
-	// The gov proposal types can be individually enabled
-	if len(enabledProposals) != 0 {
-		govRouter.AddRoute(wasmtypes.RouterKey, wasm.NewWasmProposalHandler(appKeepers.WasmKeeper, enabledProposals)) //nolint:staticcheck // we still need this despite the deprecation of the gov handler
```
dimiandre, line 612: We don't need this anymore because of sdk47 right?

Reecepbcups: yes, messages are generated & routed properly 😃

PR #874, opened 2023-10-28
Some txs are hitting FeePay when no --fee is used

Reecepbcups:
**testnet only scheduled upgrade for: v1800alpha2**

Users on testnet get

```
junod tx wasm execute juno1h2md5367062ypuv93kpwyu84eaq04xx4lfmqwqp5fkqrwa66pynsk6qmk5 {"admin":{"add_service": {"name": "rebalancer", addr": "juno18rpfddza4g3h5a05fzwq6xwepzh2t0twhetly4y5aqjyeh8cjflspa8fqr"}}} --from juno17s47ltx2hth9w5hntncv70kvyygvg0qr83zghn --gas-prices 0.025ujunox --gas auto --gas-adjustment 1.4 -y

Correct, gas attached, I used this command few days ago and it worked, 2 copies of the same contract, so I assume its the new feepay module?, no idea whats going on.
```

This will require a software upgrade on testnet as `v18` tag `.alpha.2`. Just a standard coordinated upgrade

## TODO:
-  [x] Wasm Execute, with Fee not registered (pass - this case)
- [x] wasm execute with fee & is registered (pass)
- [x] wasm exec no fee not registered (fail)
- [x] wasm exec no fee is registed (pass)


joelsmith-2019:
As of right now, FeePay considers any 0 fee, single-message Tx to be a FeePay tx. We may need to add an additional check to ensure that the client provides 0 gas OR we could simply ignore 0 fee transactions on non-registered contracts and handle them normally (with no FeePay logic interfering).

Reecepbcups:
@joelsmith-2019 what if instead of erroring, if its not a valid feePay Tx, we just run it normally with the deduct fee?

Then if deductFee fails, we error that the FeePay Tx failed and wrap it with the deductFee err?
This way even if 0 fee is added (ex: gas prices is used via CLI) it runs both cases

- check feePay Tx
- that failed, save to a var := true, try deduct fee
- that deductFee fails, show the feePay Tx and wrap deductFee err in it too.

joelsmith-2019:
@Reecepbcups, that could work, I'll give it a go

Reecepbcups:
Should bump PFM to official v7.0.1 release here too (Pigeonfall)

Reecepbcups:
GlobalFee PR just changes it to use `insufficient fees` check.

`module_globalfee_test.go:134` fails for:
```
failed to submit gov v1 proposal: exit code 1:  Error: rpc error: code = Unknown desc = rpc error: code = Unknown desc = invalid fee amount: 0ujuno: insufficient fee [CosmosContracts/juno/v18/x/feepay/ante/deduct_fee.go:226] With gas wanted: '18446744073709551615' and gas used: '9903' : unknown request
```

from the command

```
junod tx gov submit-proposal /var/cosmos-chain/juno/proposal.json --gas auto --gas-prices 0.003ujuno --gas-adjustment 1.3 --from default-juno-2-pfl --keyring-backend test --output json -y --chain-id juno-2 --home /var/cosmos-chain/juno --node tcp://juno-2-val-0-TestJunoGlobalFee:26657
```

So gas is used, but deduct fees: !fees.IsValid() is not valid since it is a 0token fee? We know this passed before on main. 
Maybe another sanity check in the `// Std sdk route` ?

joelsmith-2019:
I'll check it out and see what's up!

Reecepbcups:
works indeed

Reecepbcups:
@coderabbitai review

joelsmith-2019:
I refactored all of the wacky code from yesterday. Let me know if you like how it works and looks.

Reecepbcups (commented):
lgtm

joelsmith-2019 (approved):
LGTM

PR #875, opened 2023-10-28
v19: Enforce Validator's Max 24 Hour Change Rate To 5%

### LINKED ISSUE #819: prop 316: Update Validator's Max 24 Hour Change Rate To 5%
Per prop 316 - https://www.mintscan.io/juno/proposals/316

All validators max commission rate should be set to 5% via an upgrade (v18)
Via the ante, ensure any CreateValidator message (OR Authz nested submsg) makes a validator with a max daily channge rate >5% (or modify it so the rate is only 5%)

joelsmith-2019:
Closes #819 

- [x] Create ante handler
- [x] Unit testing (Standard Messages)
- [x] Ensure Authz inline messages are filtered
- [x] Unit testing Authz messages
- [x] Convert all validators to 5% max change in the v19 upgrade handler per [proposal](https://ipfs.io/ipfs/bafkreid26gspuwzy3wiolgue3mu3n2zxgwgckvwy4ehg55oqgwhypcn7bq) [316](https://dev.mintscan.io/juno/proposals/316)


Reecepbcups:
I wrote some Ante example test here in another repo: https://github.com/strangelove-ventures/poa/blob/399274a6c80a595f539826ff25ea64482774ef2a/ante/ante_test.go

the setBlockHeader(...) func could be modified with a set time as well. or just increase it 50k blocks every unit test

Reecepbcups:
This is going to continue as planned to v19 since it requires an Upgrade Handler to 
- iter all validators
- grab validator
- update val info MaxCommissionRate = 0.05
- save val

Reecepbcups:
This can be resumed once #906 gets in

app/ante.go
```diff
@@ -90,13 +90,14 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
...
 	anteDecorators := []sdk.AnteDecorator{
-		// GLobalFee query params for minimum fee
+		// GlobalFee query params for minimum fee
 		ante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
 		wasmkeeper.NewLimitSimulationGasDecorator(options.WasmConfig.SimulationGasLimit),
 		wasmkeeper.NewCountTXDecorator(options.TxCounterStoreKey),
 		ante.NewExtensionOptionsDecorator(options.ExtensionOptionChecker),
 		decorators.MsgFilterDecorator{},
 		ante.NewValidateBasicDecorator(),
+		decorators.NewChangeRateDecorator(&options.StakingKeeper),
```
Reecepbcups, line 100: I moved this after `NewValidateBasicDecorator` just to ensure it is a valid message before checking the change rate

PR #886, opened 2023-11-14
FeePay Documentation

joelsmith-2019:
Adds documentation for the FeePay module. Somehow that was completely overlooked.


Reecepbcups:
Nice work! just had to fix 1 minor thing with formatting on the last page

Reecepbcups (approved):

PR #887, opened 2023-11-15
fix!: Bump wasmd to v0.45

Reecepbcups:
v1800alpha3 upgrade

issues in v0.43 and 0.44 have caused many issues on testnet. The CW team has retracted these too (of which we are on testnet with) and requires a state breaking change to v0.45


joelsmith-2019 (approved):
LGTM

PR #893, opened 2023-11-22
feat!: x/clock v2 - permissionless

### LINKED ISSUE #861: x/clock - permissionless
- like feepay, allow a dev to put some funds in to pay for execution fees
- Free execution up to 100k gas, then a reduced ujuno fee compare to globalfee (ex: 0.075ujuno is current. for this maybe charge 0.00075ujuno*gas each block. something reasonable)
- if a contract execute fails enough (ex: no sudo message) "jail" the contract. small fee to unjail

joelsmith-2019:
Closes #861 

**REMAINING TASKS**

Keeper:
- [x] clock.go: Ensure senderAddress is admin (first) then creator (second)

Querier:
- [x] Get all contracts (jailed & unjailed)
- [x] Get contract by address

Genesis:
- [x] List of all registered contracts
- [x] List all jailed contracts
- [x] Display gas
    
Codec:
- [x] Update Codec: Update with new Tx Messages

EndBlocker:
- [x] Execute up to 100k gas
- [x] If contract errors, jail contract

Testing:
- [x] Unit Tests on Msg Server
- [x] Unit Tests on Querier
- [x] Unit Tests on Params (params_test.go)
- [x] Unit Tests on Genesis (genesis_test.go)
- [x] Unit Tests on Codec
- [x] Unit Tests on Msg
- [x] End Blocker Tests?
- [x] Interchain Tests
- [x] Interchain Test Unregistering a contract!

Documentation
- [x] Update Documentation with new CLI, Logic, Etc.


x/clock/abci.go
```diff
@@ -31,12 +33,19 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
...
+			err = k.JailContract(ctx, addr)
+			if err != nil {
+				log.Printf("[x/clock] Failed to Error Contract %s: %v", addr, err)
+			}
 		}
 	}
 
-	if len(errorExecs) > 0 {
+	// Log errors if present
+	if errorExists {
```
Pipello, line 48: what does this solve compared to the previous implementation `len(errorExecs) > 0`

joelsmith-2019: The length of the array was always greater than 0, so the statement was always true.

x/clock/abci.go
```diff
@@ -19,9 +19,11 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
 
 	p := k.GetParams(ctx)
 
-	errorExecs := make([]string, len(p.ContractAddresses))
+	contracts := k.GetAllContracts(ctx, false)
+	errorExecs := make([]string, len(contracts))
```
joelsmith-2019, line 23: See line 23. If more than one contract exists, the length of errorExecs will always be greater than 0.

x/clock/abci.go
```diff
@@ -31,12 +33,19 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
...
 		_, err = k.GetContractKeeper().Sudo(childCtx, contract, message)
 		if err != nil {
+			errorExists = true
 			errorExecs[idx] = addr
-			continue
+
+			// Jail contract on error
+			err = k.JailContract(ctx, addr)
+			if err != nil {
+				log.Printf("[x/clock] Failed to Error Contract %s: %v", addr, err)
```
Reecepbcups, line 42: I did this, but we should use a keeper logger instead where `logger log.Logger,` is passed into the keeper, we modify it as `logger = logger.With(log.ModuleKey, "x/"+poa.ModuleName)` and then save to the Keeper struct

```
func (k Keeper) Logger() log.Logger {
	return k.logger
}
```

then we can `k.GetLogger().Debug(...)` here

same below with `if errorExists { ...`

proto/juno/clock/v1/genesis.proto
```diff
@@ -13,17 +13,22 @@ message GenesisState {
     (gogoproto.nullable) = false,
     (gogoproto.jsontag) = "params,omitempty"
   ];
-}
-
-// Params defines the set of module parameters.
-message Params {
   // contract_addresses stores the list of executable contracts to be ticked on every block.
-  repeated string contract_addresses = 1 [    
+  repeated string contract_addresses = 2 [
```
Reecepbcups, line 17: note: this will require a module store Migration handler (to move from previous params store -> new genesis store.

Governance uploaded contracts should allow for higher gas limits.

ex: standard contract uploads only get the `Params.standard_contract_gas_limit`, but if it is uploaded via gov we set it to a higher amount (ex: Params.privileged_contract_gas_limit) that would be 100m or 2bn or something

x/clock/client/cli/tx.go
```diff
@@ -0,0 +1,129 @@
...
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+// NewUnregisterClockContract returns a CLI command handler for unregistering a
+// contract for the clock module.
+func NewUnregisterClockContract() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unregister [contract_bech32]",
+		Short: "Unregister a contract for clock. Sender must be admin of the contract.",
```
Reecepbcups, line 70: `Sender must be admin of the contract.` should be in the `Long:` cobra.Command section instead for all cmds

(`Short:` is shown in the CLI when you type in the parent ex: `junod tx clock` while `Long:` is shown when you request specific help ex: `junod tx clock mycmd --help`)

x/clock/genesis.go
```diff
@@ -35,7 +35,26 @@ func GetGenesisStateFromAppState(cdc codec.Codec, appState map[string]json.RawMe
...
 
 func ValidateGenesis(data types.GenesisState) error {
-	return data.Params.Validate()
+	err := data.Params.Validate()
+	if err != nil {
+		return err
+	}
+
+	// Validate contracts
+	for _, addr := range data.ContractAddresses {
```
Reecepbcups, line 44: With contracts moving out of the Params (since it was a gov required whitelist before), I wonder if it makes sense to not put contracts or jailed contracts into the genesis file.

instead just putting it in a `store`  and calling it a day. (not a reason to store in genesis anymore is there?)

x/clock/genesis.go
```diff
@@ -44,18 +63,36 @@ func InitGenesis(
...
 		panic(err)
 	}
 
+	// Set params
 	if err := k.SetParams(ctx, data.Params); err != nil {
 		panic(err)
 	}
+
+	// Register unjailed contracts
+	for _, addr := range data.ContractAddresses {
```
Reecepbcups, line 77: on InitGenesis there are no contracts on chain (InitGenesis is called at the start of the network, so no user interactions). See above comment related to this too

(same for jailed)

x/clock/keeper/msg_server.go
```diff
@@ -25,6 +25,57 @@ func NewMsgServerImpl(k Keeper) types.MsgServer {
 	}
 }
 
+// RegisterClockContract handles incoming transactions to register clock contracts.
+func (k msgServer) RegisterClockContract(goCtx context.Context, req *types.MsgRegisterClockContract) (*types.MsgRegisterClockContractResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Validate sender address
+	if _, err := sdk.AccAddressFromBech32(req.SenderAddress); err != nil {
```
Reecepbcups, line 33: nit: put all of these into a private method to reduce duplication for all checks

*(maybe add Validate() to MsgRegisterClockContract, MsgUnregisterClockContract, MsgUnjailClockContract if not already in msgs.go)*

x/clock/keeper/querier.go
```diff
@@ -24,13 +24,48 @@ func NewQuerier(k Keeper) Querier {
 func (q Querier) ClockContracts(stdCtx context.Context, _ *types.QueryClockContracts) (*types.QueryClockContractsResponse, error) {
 	ctx := sdk.UnwrapSDKContext(stdCtx)
 
-	p := q.keeper.GetParams(ctx)
+	contracts := q.keeper.GetAllContracts(ctx, false)
```
Reecepbcups, line 27: this likely needs to be paginated sadly. same for jailed

x/clock/keeper/querier.go
```diff
@@ -24,13 +24,48 @@ func NewQuerier(k Keeper) Querier {
...
+	ctx := sdk.UnwrapSDKContext(stdCtx)
+
+	// Ensure the contract address is valid
+	if _, err := sdk.AccAddressFromBech32(req.ContractAddress); err != nil {
+		return nil, types.ErrInvalidAddress
+	}
+
+	// Check if the contract is jailed or unjailed
+	isUnjailed := q.keeper.IsClockContract(ctx, req.ContractAddress, false)
+	isJailed := q.keeper.IsClockContract(ctx, req.ContractAddress, true)
```
Reecepbcups, line 56: this is confusing why it would check isUnjailed and isJailed? only isJailed needs checking?

x/clock/types/codec.go
```diff
@@ -30,13 +29,18 @@ func init() {
 
 // RegisterLegacyAminoCodec registers concrete types on the LegacyAmino codec
 func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
-	cdc.RegisterConcrete(Params{}, "juno/x/clock/Params", nil)
-	legacy.RegisterAminoMsg(cdc, &MsgUpdateParams{}, "juno/x/clock/MsgUpdateParams")
+	cdc.RegisterConcrete(&MsgRegisterClockContract{}, "juno/x/clock/MsgRegisterClockContract", nil)
```
Reecepbcups, line 32: these amino names are likely getting close to the limit. maybe change the namespace to just `juno/MsgRegisterClockContract` or `clock/MsgRegisterClockContract`

could be wrong but just want to call it out

x/clock/keeper/clock.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/CosmosContracts/juno/v18/x/clock/types"
+)
+
+// Store Keys for clock contracts (both jailed and unjailed)
+var (
+	StoreKeyContracts       = []byte("contracts")
+	StoreKeyJailedContracts = []byte("jailed-contracts")
```
Reecepbcups, line 13: this worries me vs a single Contracts array that stores a bool for isJailed within it. Logically a lot of true & false checks for double querying data concerns me in relation to logic issues down the line as the module expands

StoreKeyContracts = []byte("contracts")

where each contract is

```
contract {
    contract: address (string or bytes)
    isJailed: bool
}
```

joelsmith-2019: Restored to original design!

x/clock/types/codec.go
joelsmith-2019: Sounds good, prefixed with clock/

x/clock/keeper/querier.go
joelsmith-2019: Resolved with new singular store

x/clock/keeper/querier.go
joelsmith-2019: Added pagination

x/clock/client/cli/tx.go
joelsmith-2019: Thank you, updated with recent commit

x/clock/genesis.go
joelsmith-2019: Good point, they would all be jailed instantly regardless! Genesis no longer keeps track of contracts.

x/clock/genesis.go
joelsmith-2019: Contracts removed from genesis

x/clock/keeper/msg_server.go
joelsmith-2019: Great point! Just refactored the validation checks using ValidateBasic from msgs.go

x/clock/abci.go
joelsmith-2019: Implemented logger in last commit

proto/juno/clock/v1/genesis.proto
joelsmith-2019: No longer needed as no gov proposals have been made

interchaintest/helpers/clock.go
```diff
@@ -0,0 +1,120 @@
...
+	stdout, _, err := chain.Exec(ctx, cmd, nil)
+	require.NoError(t, err)
+
+	debugOutput(t, string(stdout))
+
+	err = testutil.WaitForBlocks(ctx, 1, chain)
+	require.NoError(t, err)
+}
+
+type ClockContract struct {
```
Reecepbcups, line 82: note: in the future we can use the gRPC server like I do in https://github.com/strangelove-ventures/interchaintest/pull/892 for better querying with our custom modules (and automatic type handling)

interchaintest/helpers/cosmwasm.go
```diff
@@ -38,6 +43,48 @@ func SetupContract(t *testing.T, ctx context.Context, chain *cosmos.CosmosChain,
 	return codeId, contractAddr
 }
 
+func SetupContractWithAdmin(t *testing.T, ctx context.Context, chain *cosmos.CosmosChain, creatorKeyName string, adminAddress string, fileLoc string, message string) (codeId, contract string) {
```
Reecepbcups, line 46: I am going to replace this with SetupContract with extraFlags since this is similar. Then you can just --admin in the SetupContract(...) func

x/clock/types/params_test.go
```diff
@@ -10,35 +10,54 @@ import (
...
+			types.NewParams(100),
+			false,
+		},
+		{
+			"Fail - Not Enough Gas",
+			types.NewParams(1_000),
+			false,
+		},
+		{
+			"Fail - Not Enough Gas",
```
Reecepbcups, line 48: Do / did we get a test for executing more than the block gas and ensuring they still go through?

block gas of 1m
15x contracts at 100k gas should still execute, even though they are >1m

interchaintest/helpers/clock.go
joelsmith-2019: You are referring to something along the lines of this, right?

```go
// AllBalances fetches an account address's balance for all denoms it holds
func (c *CosmosChain) AllBalances(ctx context.Context, address string) (types.Coins, error) {
	params := bankTypes.QueryAllBalancesRequest{Address: address}
	grpcAddress := c.getFullNode().hostGRPCPort
	conn, err := grpc.Dial(grpcAddress, grpc.WithTransportCredentials(insecure.NewCredentials()))
	if err != nil {
		return nil, err
	}
	defer conn.Close()

	queryClient := bankTypes.NewQueryClient(conn)
	res, err := queryClient.AllBalances(ctx, &params)

	if err != nil {
		return nil, err
	}

	return res.GetBalances(), nil
}```

File: x/clock/types/msgs.go [hunk 16]
```diff
@@ -14,23 +12,98 @@ const (
...
 
 // NewMsgUpdateParams creates new instance of MsgUpdateParams
 func NewMsgUpdateParams(
 	sender sdk.Address,
-	contracts []string,
 ) *MsgUpdateParams {
 	return &MsgUpdateParams{
 		Authority: sender.String(),
-		Params:    Params{ContractAddresses: contracts},
+		Params:    Params{},
```
Reecepbcups: Missing gasLimit

interchaintest/module_clock_test.go
```diff
@@ -33,40 +33,131 @@ func TestJunoClock(t *testing.T) {
...
+	// -- MIGRATE CONTRACT --
+	// Ensure migrations can patch contracts that error or exceed gas limit
+	// so they can be unjailed.
+	//
+
+	// Migrate the high gas contract to a contract with lower gas usage
+	helpers.MigrateContract(t, ctx, juno, user.KeyName(), contractAddr, "contracts/clock_example_migrate.wasm", `{}`)
+
+	// Unjail the contract
+	helpers.UnjailClockContract(t, ctx, juno, user, contractAddr)
```
Reecepbcups, line 95: If the contract unjails but did not migrate, the contract will remain jailed correct after the next block?
(So the flow is contractJailed -> user unjails -> EndBlock -> contractFailsAgain -> ReJail)

interchaintest/helpers/clock.go
Reecepbcups: yes, not required for this PR just in the future as a nice to have

(all that conn setup will not be required)

interchaintest/helpers/cosmwasm.go
Reecepbcups: resolved with 782ff97a1a238b9aae03a1363dd37676e724a04e

x/clock/types/msgs.go
Reecepbcups: resolved with d37317f0bca284900a005a2890c802c7f6e3f0b7

interchaintest/helpers/clock.go
Reecepbcups: yes, not needed for this PR but would be nice in the future (this is just a reference for you and me to remember)

x/clock/abci.go
```diff
@@ -1,42 +1,111 @@
...
-	if len(errorExecs) > 0 {
-		log.Printf("[x/clock] Execute Errors: %v", errorExecs)
+	// Log errors if present
+	if errorExists {
+		logger.Error("Failed to execute contracts", "contracts", errorExecs)
+	}
+}
+
+// Execute contract, recover from panic
+func executeContract(k keeper.Keeper, childCtx sdk.Context, contractAddr sdk.AccAddress, err *error) {
```
Reecepbcups, line 84: note to self

x/clock/abci_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	// Increment block height
+	s.ctx = s.ctx.WithBlockHeight(11)
+
+	return contract.ContractAddress
+}
+
+// Test the end blocker. This test registers a contract, executes it with enough gas,
+// too little gas, and also ensures the unjailing process functions.
+func (s *EndBlockerTestSuite) TestEndBlocker() {
```
Reecepbcups, line 138: Set the BlockGas (consensusParams) to < the contract executes for in the endblocker and ensure that goes through fine as well

interchaintest/module_clock_test.go
joelsmith-2019: Yes, that is correct. If the contract is not patched, it will continually get re-jailed over and over again.

interchaintest/helpers/clock.go
joelsmith-2019: Sounds good, will keep that in mind for the future!

x/clock/abci_test.go
joelsmith-2019: Line 156 updates the gas limit to 65,000 which is not enough gas for the contract. Is that what you mean?

x/clock/types/params_test.go
joelsmith-2019: We do have a performance test which runs 1000 contracts which utilize about 70k gas each in the abci_test.go (Line 2060.

Reecepbcups: perfect

x/clock/abci_test.go
Reecepbcups: I was referencing `s.ctx = s.ctx.WithBlockGasMeter(sdk.NewGasMeter(1_000_000))` instead, which is the max gas that can be used. But seeing no issues with this after I put it into test, so does not apply :D

x/clock/abci.go
```diff
@@ -1,42 +1,111 @@
...
+	}
+
+	// Track errors
+	errorExecs := make([]string, len(contracts))
+	errorExists := false
 
-	for idx, addr := range p.ContractAddresses {
-		contract, err := sdk.AccAddressFromBech32(addr)
+	// Function to handle contract execution errors. Returns true if error is present, false otherwise.
+	handleError := func(err error, idx int, contractAddress string) bool {
```
Reecepbcups, line 35: is there a way we can not have this be an anon function & make it a priv function?

re-assigning the memory every block scares me for a mem leak

joelsmith-2019: On it! The anon function was used to reduce the number of parameters we need to pass to the handleError function.

Reecepbcups (approved):

Reecepbcups:
Great work!

PR #917, opened 2023-12-16
Juno Website Hosting

Reecepbcups:
smart contract required
really rough draft, not prod ready at all

note: the current 1317 API does not allow for delivering .zip files, comment below for a working local example with a zip

Demo: https://x.com/Reecepbcups_/status/1735844874404892808?s=20

Webhost contract: https://github.com/Reecepbcups/cw-webhost


dimiandre:
Cool concept!

Reecepbcups:
The following is a local example that serves the zip files as expected. Something with the 1317 API does not like serving the images and other nested content.

With ZIP files we can embed standard CSS, JS, and images all in 1 & serve via an open port on the full node

```
package main

import (
	"archive/zip"
	"fmt"
	"io"
	"log"
	"net/http"
	"os"
	"strings"
)

func main() {
	// h := Handler()
	// http.HandleFunc("/webhost", h)
	// http.ListenAndServe(":8080", nil)

	filename := "/home/reece/Desktop/webhost/new.zip"
	if filename == "" {
		panic("need a path to a zip file")
	}
	zr, err := zip.OpenReader(filename)
	if err != nil {
		panic(err)
	}
	defer zr.Close()
	const addr = ":8080"
	log.Printf("see http://%s/", addr)

	http.ListenAndServe(addr, http.FileServer(http.FS(zr)))

}

func Handler() http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		file := "/home/reece/Desktop/new.zip"
		// file := "/home/reece/Desktop/Programming/Go/juno/app/webhost/example/example.zip"

		f, err := os.Open(file)
		if err != nil {
			panic(fmt.Sprintf("failed to open zip file %s: %s", file, err))
		}
		// size of f
		size, err := f.Stat()
		if err != nil {
			panic(fmt.Sprintf("failed to get size of zip file %s: %s", file, err))
		}

		z, err := zip.NewReader(f, size.Size())
		if err != nil {
			panic(fmt.Sprintf("failed to create zip reader for %s: %s", file, err))
		}

		// z.File[0].Name == index.html here

		// deflate-compressed files, name to index in z.File
		// m := make(map[string]int)
		for i := range z.File {

			file := z.File[i]

			// if z.File[i].Method != zip.Deflate {
			// 	// panic(fmt.Sprintf("unsupported compression method %d for %s", z.File[i].Method, z.File[i].Name))
			// 	continue
			// }
			fmt.Println(file.Method, file.Name)

			rd, err := z.File[i].OpenRaw()
			if err != nil {
				panic(fmt.Sprintf("failed to open raw file %s: %s", z.File[i].Name, err))
			}

			w.Header().Add("Vary", "Accept-Encoding")
			if z.File[i].Method == zip.Deflate {
				w.Header().Set("Content-Encoding", "deflate")
			}

			written, err := io.Copy(w, rd)
			if err != nil {
				panic(fmt.Sprintf("failed to write file %s: %s", z.File[i].Name, err))
			}

			fmt.Println("written", written)
		}

		if strings.Contains(r.Header.Get("Accept-Encoding"), "deflate") {
			// names inside zip catalog should not start with /webhost (base url from the server)
			// key := strings.TrimPrefix(r.URL.Path, "/webhost")
			key := strings.TrimPrefix(r.URL.Path, "")

			if key == "/webhost" {
				key = "index.html"
			}

			// replace any simlinks in the website index with the other content? i.e. base64 image or something?

			// if i, ok := m[key]; ok {
			// 	if rd, err := z.File[i].OpenRaw(); err == nil {
			// 		if z.File[i].Method == zip.Deflate {
			// 			w.Header().Set("Content-Encoding", "deflate")
			// 		}

			// 		io.Copy(w, rd) // sending raw compressed data
			// 		return
			// 	}
			// }
		}

		// write the website contents to the user

		// w.Header().Set("Content-Type", "text/html; charset=utf-8")
		// w.WriteHeader(http.StatusOK)
		// w.

		// srv.ServeHTTP(w, r) // fallback

		// return w and r to the user

		// srv.ServeHTTP(w, r)

		// writeResponse(w, "hello world "+z.File[0].Name)
	}

}
```

faddat:
this is sensible and good and Juno herself should host her website.  Not others.

Reecepbcups:
Charter people are against feature upgrades. It will not be getting in.

PR #940, opened 2024-01-13
chore: update README.md

sqltrigger:
Came across a dead link while reading it. It appears ROADMAP.md has been deleted in #811.
Removed it to avoid a File Not Found link.
Hope I could be of any help. Have a nice evening.


dimiandre (approved):
thanks!

PR #952, opened 2024-01-25
fix(upgrade): clock params

Reecepbcups:
Forgot to put clock params in the upgrade handler

testnet: v1900alpha2 upgrade
mainnet: backported same change from `v1900alpha2`

Will go live on testnet next week, confirm clock is good, then mainnet


Reecepbcups:
- Prop up tomorrow morning (Jan 26th) for alpha upgrade
- Live on Uni Jan 29th 1600UTC
- Confirm working
- Tag v19 official Jan 29th
- push on chain
- live Feb 5th mainnet

joelsmith-2019 (approved):
Good catch!

PR #959, opened 2024-02-05
patch: Migrate the correct amount of Core-1 funds

Reecepbcups:


joelsmith-2019 (approved):

Reecepbcups:
hardcode breaks the v19 test but does not matter since it wont be doing again

kopeboy:
I see the hardcoded number is different than the one in the comment (which has one more digit, 5, at the end): which one is right? @joelsmith-2019 did you review it?

Reecepbcups:
@kopeboy the 5 is supposed to be at the end yes. miscopy on my end so only 530k got transferred instead of the 5.3m

PR #960, opened 2024-02-05
fix!: v19 not loading in contracts from state

Reecepbcups:
## Timeline:
- uni-6 had an issue when I was using 2 different wasmvms (one for the new light client, one for standard wasm)
- I called `InitializePinnedCodes` on both, which resulted in issues since it was double pinning codes
- To resolve this, I switched to use 2 different vms
`mainWasmer, err := wasmvm.NewVM(wasmDir, wasmCapabilities, 32, wasmConfig.ContractDebugMode, wasmConfig.MemoryCacheSize)` & `lcWasmer, err := wasmvm.NewVM(lcWasmDir, wasmCapabilities, 32, wasmConfig.ContractDebugMode, wasmConfig.MemoryCacheSize)` in keepers/keepers.go
- I forgot to re-add the InitializePinnedCodes for wasm, since previously the wasmlc was handling the loading into cache

## Also needs a backport to main
- And fix testing

PR #961, opened 2024-02-05
fix! not loading in contracts with custom WasmVM after upgrade

Reecepbcups:
ref PR #960 for further information and debugging scripts

Will need to backport to main

- Tag as v19.2? (or v20 since it is state breaking, just with the v19 namespace)
- need to try on testnet
- Need to halt and patch to mainnet vals


Reecepbcups:
@dimiandre I'm fine with that. Didn't know if Core-1 could get the votes. 

The first commit in this PR is from the release/v19 branch. So I can't remove that but will get the handler out if desired

Reecepbcups:
lets do the halt and patch

Reecepbcups:
Can confirm v18 -> this branch works as expected with it.
image

dimiandre (approved):
looks good

PR #993, opened 2024-04-09
chore!: v22 Namespace & Upgrade Handler

dimiandre:
1. v22 upgrade handler
2. v22 namespace
3. version bumps - including ibc-go


dimiandre:
seems tests are broken because of 
https://github.com/strangelove-ventures/interchaintest/pull/1043

thanks @faddat for pointing out

Reecepbcups:
Comet had a minor state breaking change (types). Can use `github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.37.2` as well since interchaintest does not depend on actual cometbft, but rather just the types and func signatures

dimiandre:
> Comet had a minor state breaking change (types). Can use `github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.37.2` in the `interchaintest/` go.mod as well since interchaintest does not depend on actual cometbft, but rather just the types and func signatures

thanks reece! seems to do the trick!

dimiandre:
> Are the version bumps needed for this specific chain upgrade or can we do them later, eg. with the help of a new maintainer from Dev Dept's Blockchain engineer RFP? (ofc help by anyone in the community is always appreciated and could be retroactively compensated).
> 
> I see a lot of errors (in 8 out of 9 x/modules + in app/apptesting/test_suite.go and app/test_helpers.go, etc.) are caused by the missing functions `RequestBeginBlock` & `RequestEndBlock` in the new **cometbft**. We are bumping it from 0.37.2 to 0.38.4, but they were removed starting from `v0.38.0`, and I'm a noob but it doesn't seem required: image
> 
> EDIT: I had this comment ready but I posted it late 😅 Happy that you already figured the issue out and that I'm learning to review PRs here 😇

version bumps here are minor, and needed to solve the security issues we had with ibc-go. 
When you tidy go deps the versions get minor-bumped automatically

Also ibc-go v7.4 is state breaking comparing v7.3 that's why i'm cutting the v22 release

docker-compose.yml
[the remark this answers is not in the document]
dimiandre: No, that is used only for local tests. 

Nothing to do with production builds

docker-compose.yml
```diff
@@ -15,3 +15,4 @@ services:
       - GAS_LIMIT=${GAS_LIMIT:-10000000}
       - STAKE_TOKEN=${STAKE_TOKEN:-ujunox}
       - TIMEOUT_COMMIT=${TIMEOUT_COMMIT:-5s}
+      - UNSAFE_CORS=true
```
kopeboy, line 18: I guess we'll need to remember to disable UNSAFE_CORS before merging into a release, right?

interchaintest/go.mod
```diff
@@ -8,19 +8,20 @@ replace (
 	github.com/ChainSafe/go-schnorrkel => github.com/ChainSafe/go-schnorrkel v0.0.0-20200405005733-88cbf1b4c40d
 	github.com/ChainSafe/go-schnorrkel/1 => github.com/ChainSafe/go-schnorrkel v1.0.0
 	// For this nested module, you always want to replace the parent reference with the current worktree.
-	github.com/CosmosContracts/juno/v21 v21.0.0-00010101000000-000000000000 => ../
+	github.com/CosmosContracts/juno/v22 v22.0.0-00010101000000-000000000000 => ../
 	github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.2 //indirect
+	github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.37.2
```
kopeboy, line 13: Shouldn't this be `v0.37.4`?

kopeboy (approved):

faddat:
Every time Juno is upgraded, best practices are:

* use the latest major golang release
* use the latest release of EVERY cosmos ecosystem package

The reason for this is that there are frequently security improvements, and performance improvements.  I can't think of a time when it was a bad thing to use the latest dependencies.  It is reasonable to use cosmos-sdk and ibc-go as a guide for what's freshest.

faddat (changes requested):
This doesn't update some ecosystem libraries.  I've done the updates, you'd just need to merge the PR below:

https://github.com/CosmosContracts/juno/pull/994

x/tokenfactory/keeper/bankactions.go
```diff
@@ -68,6 +72,28 @@ func (k Keeper) forceTransfer(ctx sdk.Context, amount sdk.Coin, fromAddr string,
...
+	fromAcc, err := sdk.AccAddressFromBech32(fromAddr)
+	if err != nil {
+		return err
+	}
+
+	sortedPermAddrs := make([]string, 0, len(k.permAddrs))
+	for moduleName := range k.permAddrs {
+		sortedPermAddrs = append(sortedPermAddrs, moduleName)
+	}
+	sort.Strings(sortedPermAddrs)
```
Reecepbcups, line 84: is this deterministic? I know sort.Slice is not, I assume it's the same here?

If you can't sort, iter through all, don't break early. Then if an error, return after the for loop

x/tokenfactory/keeper/bankactions.go
```diff
@@ -68,6 +72,28 @@ func (k Keeper) forceTransfer(ctx sdk.Context, amount sdk.Coin, fromAddr string,
...
+
+	for _, moduleName := range sortedPermAddrs {
+		account := k.accountKeeper.GetModuleAccount(ctx, moduleName)
+		if account == nil {
+			return status.Errorf(codes.NotFound, "account %s not found", moduleName)
+		}
+
+		if account.GetAddress().Equals(fromAcc) {
+			return status.Errorf(codes.Internal, "send from module acc not available")
```
Reecepbcups, line 93: How would someone send from a module account? do you mean toAcc? not following here.

if there is an issue I would like to get it upstreamed to my own widely used fork of TF but I don't see the case here

JakeHartnell (approved):

PR #994, opened 2024-04-11
juno v22 with ecosystem updates

faddat:
This PR updates the juno v22 release so that it uses fresh cosmos ecosystem libraries.

* comet will have the latest improvements
* cosmos-sdk will be updated to solve a security issue
* latest math and errors libraries will be used

Don't make new releases without applying the latest patches.

Process: visit the repository of each dependency and check to see if Juno is using the latest version of the dependency in question.

Using the latest version of Go makes it impossible for the validator set to have mixed runtimes.


dimiandre:
something changed in app.toml? 

```
        	            	9:01AM ERR failure when running app err="failed to parse /var/cosmos-chain/juno/config/app.toml: '' expected a map, got 'interface'"

```

faddat:
Maybe.  Just, it seems to break only one thing, interchaintest.

dimiandre:
> I have no idea how to fix these "login denied" things tho. good luck. Should move hosting of images and stuff off of heighliner and onto ghcr.io

When doing PRs is always best practice making sure you are not breaking other stuff,

> Maybe. Just, it seems to break only one thing, interchaintest.
> 
> Would be nice to be able to keep Juno up to date. Check the changelogs for the things that I bumped in addition to the things you bumped. The fixes are nontrivial.
> 
> Recommend that you review my changes once again, and ask yourself:
> 
> * should this break ict?
> * does this break ict?

It's not only interchain test, this PR breaks juno init on standard builds, docker and anywhere else. 

```
git clone 
make install

junod init moniker --chain-id test

11:52AM ERR failure when running app err="failed to parse /Users/dimi/.juno/config/app.toml: '' expected a map, got 'interface'"
```

I'm on it, next time would be better to do complete PRs not just version bumps without even checking if node starts after it

Reecepbcups:
> edit: it's viper. 1.18 doesn't work with sdk 47.10 InterceptConfigsPreRunHandler

Correct, it's an SDK issue. This fixes the SDK

```go.mod
// breaks SDK app.toml parsing in ictest.
replace github.com/spf13/viper => github.com/spf13/viper v1.17.0
```

faddat:
would https://github.com/strangelove-ventures/interchaintest/pull/1043

fix the parsing?

dimiandre:
viper issue seems to be fixed with sdk v0.47.11

since jacob already bumped it i'll merge this PR. We'll have to bump it again once the v0.47.11 release is officially out

thanks all!

PR #1089, opened 2025-01-28
v28: Cosmos SDK 0.50, Wasm 2.0, Refactors, Cleanup

vexxvakan:
# v28

### Tools
- Added mise-en-place for easy version management
- Added Swagger client files
- Updated Dockerfile
- Cleaned up Buf & Makefile
- Fixed Swagger generation

### Changes
- Bumped package version
- Upgraded Cosmos SDK to v0.50.11
- Upgraded to CosmWasm v2 with wasmvm v2.2.2
- Upgraded ibc-go to v8.5.3
- Upgraded Comet to v0.38.17
- Updated ALL modules to modern implementations (more to to do later, collections migration)
- Updated and cleaned up ALL Protobuffers
- Generated API
- Removed deprecated POB & x/builder module
- Removed 08-wasm light client
- Implemented auto-cli
- Extracted wasmbindings


dimiandre (approved):

cmd/junod/cmd/root.go
```diff
@@ -1,63 +1,100 @@
...
 )
 
 // NewRootCmd creates a new root command for junod. It is called once in the
 // main function.
-func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
-	encodingConfig := app.MakeEncodingConfig()
-
+func NewRootCmd() *cobra.Command {
+	tempDir := tempDir()
+	sdk.DefaultBondDenom = "ujuno"
```
niilptr, line 67: Hi! It would be better imo to keep app.params (/app/params/const.go) and use these where needed (here and in Bech32Prefix).
@dimiandre what do you think about it?

vexxvakan: i agree this is currently pretty hardcoded but i tried cleaning up as much folder and files as possible where only one single struct or func was declared - I can however make this not hardcoded by default for sure!

niilptr (approved):

cmd/junod/cmd/root.go
vexxvakan: close: moved to v29 because its not important

PR #1125, opened 2025-02-24
fix wasmdir typo

dimiandre:
## Summary by CodeRabbit

- **Refactor**
  - Revised the underlying directory configuration to use a unified data path for enhanced consistency, setting the stage for smoother future enhancements.

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -219,7 +219,6 @@ func NewAppKeepers(
 	ac := authcodec.NewBech32Codec(bech32Prefix)
 	invCheckPeriod := cast.ToUint(appOpts.Get(server.FlagInvCheckPeriod))
 	dataDir := filepath.Join(homePath, "data")
-	wasmDir := filepath.Join(dataDir, "wasm")
 
 	// set the BaseApp's parameter store
 	appKeepers.ConsensusParamsKeeper = consensusparamkeeper.NewKeeper(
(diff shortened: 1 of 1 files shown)


app/keepers/keepers.go
```diff
@@ -549,7 +548,7 @@ func NewAppKeepers(
 		appKeepers.TransferKeeper,
 		bApp.MsgServiceRouter(),
 		bApp.GRPCQueryRouter(),
-		wasmDir,
+		dataDir,
```
Reecepbcups, line 551: be careful with this. it could result in what cause the failure in v19 with the 'technically successful' except for current contracts bug. unless you also write a migration to move all the previous data dir -> to the new reference

ref: https://medium.com/@reecepbcups/juno-v19-post-mortum-969f251bffb7

dimiandre: Hi Reece!

Thanks for the heads up, path it's indeed different from 27.0.0 

I'll double check again, it seems to be working on uni 🤔

vexxvakan: Thank you @Reecepbcups, i accidentally changed the wasmKeeper to use wasmDir in v28.0.0 but it was dataDir before that on v27.0.0 so the contracts from before the sdk 0.50 upgrade should still be in place... as long as no one instantiated or stored a code with v28 so far? If there was anything new to wasm state from v28 on it might be a problem

In #1131 I readded the custom wasmer instance you created for the wasm light client setup before and now the directory structure is exactly the same as it was previously so i believe this would work

Reecepbcups: 👍🏻

vexxvakan:
fixes are included in #1141 because of a merge delay from #1131 - changes already released as v28.0.1

PR #1203, opened 2026-07-02
fix(x/drip): eliminate duplicate AccAddressFromBech32 call and fix re…

reecepbcups:
…dundant error wrap

DistributeTokens called sdk.AccAddressFromBech32 twice on the same address: once to validate (discarding the result with _) and again after the authorization check to actually use the value. The second parse is redundant — if the first succeeded, the second always will too.

Also fixed the error message on the failed parse: the original used errorsmod.Wrapf(err, "invalid sender address: %s", err.Error()) which embeds err.Error() in the format string *and* wraps err, causing the underlying error to appear twice in the error chain. Changed to errorsmod.Wrapf(err, "invalid sender address %s", msg.SenderAddress) which includes the address for context without duplicating the error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sender address validation during token distribution.
  * Error messages now preserve the original sender address when parsing fails.
  * Reduced duplicate address parsing and reused the validated sender for the transfer step.

### DIFF
--- a/x/drip/keeper/msg_server.go
+++ b/x/drip/keeper/msg_server.go
@@ -38,8 +38,9 @@ func (ms msgServer) DistributeTokens(
 		return nil, errors.New("sender address cannot be empty")
 	}
 
-	if _, err := sdk.AccAddressFromBech32(msg.SenderAddress); err != nil {
-		return nil, errorsmod.Wrapf(err, "invalid sender address: %s", err.Error())
+	sender, err := sdk.AccAddressFromBech32(msg.SenderAddress)
+	if err != nil {
+		return nil, errorsmod.Wrapf(err, "invalid sender address %s", msg.SenderAddress)
 	}
 
 	if msg.Amount == nil || msg.Amount.Empty() {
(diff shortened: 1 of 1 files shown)


x/drip/keeper/msg_server.go
```diff
@@ -38,8 +38,9 @@ func (ms msgServer) DistributeTokens(
 		return nil, errors.New("sender address cannot be empty")
 	}
 
-	if _, err := sdk.AccAddressFromBech32(msg.SenderAddress); err != nil {
-		return nil, errorsmod.Wrapf(err, "invalid sender address: %s", err.Error())
+	sender, err := sdk.AccAddressFromBech32(msg.SenderAddress)
+	if err != nil {
+		return nil, errorsmod.Wrapf(err, "invalid sender address %s", msg.SenderAddress)
```
vexxvakan, line 43: would probs be better if this was a x/drip/types/errors.go defined error or potentially use [ErrInvalidAddress](https://github.com/cosmos/cosmos-sdk/blob/8c5746d4b1a0c5c28a3f3b844e68c000aabe775f/types/errors/errors.go#L28) from errorsmod itself

vexxvakan (approved):
Not sure if you wanna do the cleaner error path, i'll approve anyway

x/drip/keeper/msg_server.go
reecepbcups: oh this is just some juno bot I setup because @JakeHartnell sent me some stuff on it. I decided to have it do a PR on its own and it found this

